PR #6, opened 2019-02-07
Updated Pulumi CLI Version

coreycoto:


andy9775:
0.16.16 available now

chrsmith:
I don't think updating the `Dockerfile` with a newer version of the CLI is the right call here. Since it would potentially introduce a behavior change to any repos that depend on the `pulumi/actions` container today.

I filed https://github.com/pulumi/actions/issues/10 to track changing the default behavior and supporting a way to make older CLI versions available, so that we can have the common case (e.g. `:latest`) always pick up the latest stable CLI release, while still allowing developers to choose a version-specific container image if needed.

chrsmith:
So the linked issue, https://github.com/pulumi/actions/issues/10, has been fixed. The next Pulumi SDK release will also publish the `pulumi/actions` container.

`pulumi/actions:latest` will pull the latest version of the Pulumi CLI. `pulumi/actions:vX.Y.Z` will pull a specific version, in case you don't want to grab the latest.

I'll try to publish an update to `pulumi/actions:latest` with the current version of the CLI tomorrow, so that you don't need to wait until the next full SDK release to use newer CLIs for GitHub Actions.

PR #39, opened 2020-12-09
Add yarn workspace as configuration option.

grahamplata:


chrsmith (commented):
I'll wait for @stack72 to leave the final approval, but can we come up with a clearer environment variable name?

If I understand this correctly, the value of `USE_YARN_WORKSPACE` should be the name of the `yarn` workspace to use, right? Is that a common convention? `USE_FOO` should contain the name of the "foo" to use? ... or would `YARN_WORKSPACE_NAME` or simply `YARN_WORKSPACE` be more appropriate? (And like you have in the Bash script, just checking if it has a non-zero value before running `yarn workspace ${VAR} install`.

grahamplata:
Correct, and I agree with 
> YARN_WORKSPACE_NAME or simply YARN_WORKSPACE be more appropriate?
I can make those adjustments shortly.
Let me confirm the non-zero value before moving on.

stack72 (approved):

stack72:
Thanks so much for this @grahamplata

grahamplata:
No problem! Thanks!

PR #46, opened 2021-01-20
feat: Add configuration

### LINKED ISSUE #45: (v2) Configuration handling
In the current `action.yml` file, we have configured the following:

https://github.com/pulumi/actions/blob/25b22d49a02b9fec359213f05a4751c6a2d71393/action.yaml#L7-L13

(Don't mind the typo 😅)

Do we need more variables?

I am thinking about using [runtypes](https://github.com/pelotom/runtypes) to determine if our configuration is valid. Any thoughts?

cobraz:
This adds configuration to our action step.

There's one small catch, we are evaluating the input variables multiple times. In `action.yml` and on every input which is required we validate as well, the problem is that we don't get types based on either of these validations and generally I don't like typecasting.

`runtypes` offers more than what the Github Actions configuration does as well.

fixes #45

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,0 +1,22 @@
+import { getInput } from '@actions/core';
+import * as rt from 'runtypes';
+
+export const command = rt.Union(
+  rt.Literal('up'),
+  rt.Literal('refresh'),
+  rt.Literal('destroy'),
+);
+
+export const config = rt.Record({
+  command: command,
+  stackName: rt.String,
+});
+
+export type Config = rt.Static<typeof config>;
+
+export async function makeConfig(): Promise<Config> {
+  return config.check({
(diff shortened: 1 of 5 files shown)


src/config.ts
```diff
@@ -0,0 +1,22 @@
+import { getInput } from '@actions/core';
+import * as rt from 'runtypes';
+
+export const command = rt.Union(
```
cobraz, line 4: @stack72 Should we restrict to only allow these or more?

stack72: we also need to add support for preview

cobraz: Great, adding!

stack72 (approved):

src/config.ts
cobraz: Fixed! :)

PR #50, opened 2021-01-30
feat: Add basic implementation

cobraz:
This pull request adds basic implementation, which means it works.

- Adds `envalid` to validate required environmental variables.
- Uses Pulumi Automation API to select local stack
- Runs `preview`, `up`, `refresh` or `destroy`

**PS**: ~~There are 406 files changed in this repository, which is caused by the fact that our building tools result in appending a lot of extra files.~~ removed for clearing up diff.


stack72:
Can we ignore the typescript folder for now?

cobraz:
Yes, give me a second :)

.gitattributes
```diff
@@ -0,0 +1 @@
+dist/** linguist-generated
```
cobraz, line 1: Added this for convenience.

package.json
```diff
@@ -6,7 +6,8 @@
   "license": "Apache-2.0",
   "scripts": {
     "clean": "rimraf coverage build tmp dist",
-    "build": "ncc build src/main.ts -o dist --source-map --license LICENSE",
+    "prebuild": "yarn run clean",
+    "build": "ncc build src/main.ts -e typescript -o dist --source-map",
```
cobraz, line 10: `-e typescript` removes the typescript from `dist/`

src/libs/utils.ts
```diff
@@ -1,8 +1,11 @@
+import * as core from '@actions/core';
 export function invariant(
   condition: any,
   message?: string,
 ): asserts condition {
   if (!condition) {
-    throw new Error(message);
+    core.setFailed(message);
+    process.exit();
+    // throw new Error();
```
cobraz, line 9: Don't know if this is the way we'd like to do it, or if we want to throw and catch it somewhere else?

package.json
cobraz: @stack72 When using `-e typescript`, the workflow fails.

```shell
Error: Cannot find module 'typescript'
Require stack:
- /home/runner/work/_actions/pulumi/actions/add-basics/dist/index.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:797:15)
    at Function.Module._load (internal/modules/cjs/loader.js:690:27)
    at Module.require (internal/modules/cjs/loader.js:852:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.5034 (/home/runner/work/_actions/pulumi/actions/add-basics/webpack:/pulumi-github-action/external "typescript":1:1)
    at __nccwpck_require__ (/home/runner/work/_actions/pulumi/actions/add-basics/webpack:/pulumi-github-action/webpack/bootstrap:20:1)
    at Object.8338 (/home/runner/work/_actions/pulumi/actions/add-basics/webpack:/pulumi-github-action/node_modules/@pulumi/pulumi/runtime/closure/parseFunction.js:16:1)
    at __nccwpck_require__ (/home/runner/work/_actions/pulumi/actions/add-basics/webpack:/pulumi-github-action/webpack/bootstrap:20:1)
    at Object.1260 (/home/runner/work/_actions/pulumi/actions/add-basics/dist/index.js:35176:25)
    at __nccwpck_require__ (/home/runner/work/_actions/pulumi/actions/add-basics/webpack:/pulumi-github-action/webpack/bootstrap:20:1)
```

cobraz: https://github.com/vercel/ncc/issues/642

stack72: So the typescript parameter is not required? If so this is a great news!

src/libs/utils.ts
stack72: I think this is much more in keeping with how actions should be written - right?

stack72 (approved):

stack72:
Left a few small comments but this is a fantastic way to get started

src/libs/utils.ts
cobraz: Yes, but considering this, we could do it the other way too.

https://github.com/pulumi/actions/blob/a7bbd3813555fb55dd615ed5b07c3f6e3cde0b90/src/main.ts#L50-L58

I don't know. We can probably come back to this.

PR #66, opened 2021-02-22
Adding a python stack as a test workflow

stack72:


README.md
```diff
@@ -24,47 +18,37 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
```
cobraz, line 20:
```suggestion
      - uses: actions/checkout@v2
      - uses: pulumi/action-install-pulumi-cli@v1
```

Probably want that too :)

README.md
```diff
@@ -179,27 +222,24 @@ jobs:
...
-      - uses: pulumi/actions@v1
+      - name: Setup DotNet
+        uses: actions/setup-dotnet@v1
+        with:
+          dotnet-version: 3.1
```
cobraz, line 228:
```suggestion
      - name: Setup Node
        uses: actions/setup-node@v1
        with:
          node-version: 14.x
```

README.md
```diff
@@ -179,27 +222,24 @@ jobs:
...
+      - name: Install pulumi
```
cobraz, line 229:
```suggestion
      - uses: google-github-actions/setup-gcloud@master
        with:
          service_account_key: ${{ secrets.GCP_KEY }}
          project_id: ${{ env.PROJECT_ID }}
          export_default_credentials: true
      - name: Install pulumi
```

cobraz (approved):
Looks good! 👍 I think maybe we should move all the examples into a `docs/` directory?

stack72:
> Looks good! 👍 I think maybe we should move all the examples into a `docs/` directory?

why does rather than the README? I may be missing some convention here of course

.github/workflows/test.yml
```diff
@@ -44,3 +44,97 @@ jobs:
...
+      - uses: ./
+        if: always()
+        env:
+          PULUMI_CONFIG_PASSPHRASE: not-a-secret
+        with:
+          command: ${{ matrix.command }}
+          cloud-url: file://~
+          stack-name: dev
+          work-dir: .github/test-stacks/python
+  test-workflow-dotnet: # make sure the action works on a clean machine without building
```
cobraz, line 79: We should probably add the different languages into each workflow file.

cobraz:
> > Looks good! 👍 I think maybe we should move all the examples into a `docs/` directory?
> 
> why does rather than the README? I may be missing some convention here of course

Just personal preference. :)

README.md
```diff
@@ -24,47 +18,111 @@ jobs:
...
 
-## Configuring pulumi-action
+* `work-dir` (optional) - The location of your Pulumi files. Defaults to `./`.
 
-The pulumi-action can be configured in 2 ways:
+* `cloud-url` - (optional) - the Pulumi backend to login to. This would be the equivalent of what would be passed to the
+  `pulumi login` command. The action will login to the appropriate backend on your behalf provided it is configured with
+  the correct access credentials for that backend.
+  
+* `comment-on-pr` - (optional) If specified, then the action will add the results of the Pulumi action to the PR
```
cobraz, line 52:
```suggestion
* `comment-on-pr` - (optional) If `true`, then the action will add the results of the Pulumi action to the PR
```

stack72: Nice catch

stack72:
> > > Looks good! 👍 I think maybe we should move all the examples into a `docs/` directory?
> > 
> > 
> > why does rather than the README? I may be missing some convention here of course
> 
> Just personal preference. :)

let's follow up with this one in our status call - I want to make sure we get this how you think would be best

PR #68, opened 2021-02-22
feat: Add outputs

cobraz:
This pull request adds the ability to retrieve both the logs, so it can be sent to another workflow step (for a more custom comment on PR setup maybe or a Slack bot?).

It also adds all exported outputs from Pulumi to Github Workflow outputs. This is probably the most exciting part. Also, we mask outputs that are marked as secrets 👍


stack72:
Again, it would be awesome to see this in action from the test stack :)

stack72 (approved):

PR #70, opened 2021-02-22
chore: Improve jest coverage reporting

cobraz:


stack72 (approved):
This is way better :)

PR #72, opened 2021-02-24
Adding dependabot to ensure we can check for latest npm and github actions

stack72:


cobraz (approved):
Nice that you added `github-actions`! 👍

PR #82, opened 2021-02-24
Starting to build a CI matrix for the different action workflows

stack72:
@cobraz this adds:

* Separated workflows for each runtime that run all commands on linux, macos and windows
  * Please note I had to change the RandomPet example as that started throwing a panic on the CLI....
* Added a smoke test script for a specific CLI version as part of pulumi/pulumi releases
* Removed the work to output all outputs as it was just picking up the "Updating dev..." - I think there's something incorrect there around the map


stack72:
Self merging to get a green build - we can retrofit any PR changes

PR #85, opened 2021-02-26
Adding a migration guide to the README

stack72:


cobraz (approved):

action.yml
```diff
@@ -1,4 +1,4 @@
-name: Pulumi
```
cobraz, line 1: Nice catch 👍

PR #96, opened 2021-03-01
Remove AWS env from GCP example

grahamplata:
Remove AWS env from GCP example


stack72:
Thanks so much for finding this - it's very confusing for sure!

grahamplata:
No worries! Happy to help! I just realized it may need something along these lines instead:
 ```
...
env:
  GOOGLE_CREDENTIALS: ${{ secrets.GOOGLE_CREDENTIALS }}`
```

PR #117, opened 2021-03-16
feat: Add extra options

### LINKED ISSUE #88: Passing arguments (e.g. `diff`, `parallel`)
Currently, there is no way to pass any arguments. During development, we added this input:
https://github.com/pulumi/actions/blob/f1f4ffe44d9a9375fc3196af16f72a8e12b1b265/src/config.ts#L21

Due to the way Pulumi Automation API is built and the way they are handled, we can't just pass that 👆 over to any of our functions and call it a day. We either have to add some functionality in order to pass arguments to Automation API or find another option.

As of right now, Automation API requires `pulumi` CLI. This might change in the future, so IMHO, we should probably not expose arguments as if they were command-line arguments/parameters (e.g. `--diff`). We should expose them as named arguments (e.g. `diff: `), so they can be passed along to Automation API.

We should also update our README to explain that this action now uses Pulumi Automation API.

cobraz:
fixes #88

**Note**: `args` is removed.


.eslintrc.json
```diff
@@ -4,8 +4,7 @@
     "plugin:@typescript-eslint/eslint-recommended",
     "plugin:@typescript-eslint/recommended",
     "plugin:jest/recommended",
-    "prettier",
-    "prettier/@typescript-eslint"
+    "prettier"
```
cobraz, line 7: "prettier/@typescript-eslint" has been merged into "prettier" in eslint-config-prettier 8.0.0. See: https://github.com/prettier/eslint-config-prettier/blob/main/CHANGELOG.md#version-800-2021-02-21

.github/workflows/workflow_arguments.yml
```diff
@@ -0,0 +1,45 @@
...
+          PULUMI_CONFIG_PASSPHRASE: not-a-secret
+        with:
+          command: ${{ matrix.command }}
+          cloud-url: file://~
+          stack-name: dev
+          work-dir: .github/test-stacks/nodejs
+          parallel: 3
+          message: this-is-just-a-test
+          expect-no-changes: false
+          diff: true
```
cobraz, line 45: Missing `target` and `targetDependents` due to the small test stack 🤷‍♂️

README.md
```diff
@@ -53,12 +53,35 @@ The action can be configured with the following arguments:
...
   will login to the appropriate backend on your behalf provided it is configured
   with the correct access credentials for that backend.
+
 - `comment-on-pr` - (optional) If `true`, then the action will add the results
   of the Pulumi action to the PR
 
 - `github-token` - (required if comment-on-pr) A GitHub token that has access
   levels to allow the Action to comment on a PR.
 
+### Extra options
```
stack72, line 63: We should add a small comment here to suggest that these extra options, are the equivalent of the command line args

action.yml
```diff
@@ -27,6 +27,30 @@ inputs:
     description: 'A cloud URL to log in to'
     required: false
     default: ''
+  parallel:
+    description: 'Allow P resource operations to run in parallel at once (1 for no parallelism). Defaults to unbounded.'
+    required: false
+    default: '2147483647'
```
stack72, line 33: this relates to what the automation api does - this isn't actually the number we pass, this is to document what we can do

stack72 (approved):

PR #118, opened 2021-03-16
feat: Add `upsert` option

### LINKED ISSUE #9: Add support for Review Pulumi Stacks
Allow the ability for Pulumi stack to be automatically created / destroyed with each PR. This would be a super cool feature to dynamically test preview links for infrastructure changes without having to merge to staging branch.

### LINKED ISSUE #90: Creating stacks if they don't exist
Pulumi Automation API supports creating stacks if they don't exist. I think it could be useful to make this possible in the Github Action too.

With it, users can build a workflow that does not require developers/contributors of a repository to have Pulumi installed for all operations. For example, if you want to spin up a stack (environment), for testing.

Fixes: #9

cobraz:
Creates stacks if they don't exists when `upsert` is set to `true`.

fixes #90
fixes #9

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -19,6 +19,7 @@ export const config = rt.Record({
   githubToken: rt.String.Or(rt.Undefined),
   commentOnPr: rt.Boolean,
   args: rt.String.Or(rt.Undefined),
+  upsert: rt.Boolean.Or(rt.Undefined),
 });
 
 export type Config = rt.Static<typeof config>;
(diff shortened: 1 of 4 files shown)


src/config.ts
```diff
@@ -33,5 +34,6 @@ export async function makeConfig(): Promise<Config> {
     githubToken: getInput('github-token'),
     commentOnPr: getInput('comment-on-pr') === 'true' ? true : false,
     args: getInput('args') || args.join(' '),
+    upsert: getInput('upsert') === 'true' ? true : false,
```
stack72, line 37: "false" by default to ensure we maintain backwards compatibility

stack72 (approved):

PR #119, opened 2021-03-16
ci: Automatically update dist files on every push

### LINKED ISSUE #113: Don't accept built `dist/` files
As it is very easy to overlook changes in `dist/`, people can more easy send pull requests where they put malicious code in there. We should find a solution where this is either solved by maintainers or by a GHA workflow.

### Potential tasks

- [ ] Remove `yarn build` from `husky`
- [ ] Add a GHA step that builds `dist` files on every Pull Request or push to `master` (potentially pushes back to the branch?)

/cc @stack72

cobraz:
fixes #113

The @pulumi-bot comment below is just an example of what would happen if a user creates a pull request where files in `dist/` are changed. Would love some opinions on whether we want that comment and what we should write :)

/cc @infin8x @stack72

### DIFF
--- a/.github/husky/pre-commit
+++ b/.github/husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn build && git add dist/* && yarn lint-staged
+yarn lint-staged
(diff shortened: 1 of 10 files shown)


stack72 (approved):
We should just add to the README that the user doesn't need to do anything with the dist/ folder as this is handled internally

stack72:
We will need to ensure that yarn build + yarn install is added as part of every test step

PR #128, opened 2021-03-20
feat: Refresh

### LINKED ISSUE #126: Support for `refresh`
Some of our workflows used the `--refresh` arg in `v1` of the Pulumi GHA. We've successfully migrated to `v2` now, but with some _hacks_. Such as how we refresh our stack.

Our workflow looks like this:

```yaml
- uses: pulumi/actions@v2
  with:
    command: refresh
    cloud-url: gs://folio-accounting-integrations-pulumi
    stack-name: accounting-integrations-${{ github.event.inputs.environment }}
    work-dir: packages/accounting-integrations
    github-token: ${{ secrets.GITHUB_TOKEN }}

- uses: pulumi/actions@v2
  with:
    command: up
    cloud-url: gs://folio-accounting-integrations-pulumi
    stack-name: accounting-integrations-${{ github.event.inputs.environment }}
    work-dir: packages/accounting-integrations
    github-token: ${{ secrets.GITHUB_TOKEN }} 
```

What we want is to pass an input, (e.g. `refresh: true`), which does the same as `--refresh` did for us. AFAIK, we either have to do `stack.refresh()` or ask to have this functionality added to Pulumi Automation SDK.

/cc @stack72 @komalali

cobraz:
fixes #126

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -37,12 +37,18 @@ const main = async () => {
     ? LocalWorkspace.createOrSelectStack(stackArgs)
     : LocalWorkspace.selectStack(stackArgs));
 
-  core.startGroup(`pulumi ${config.command} on ${config.stackName}`);
-
   const onOutput = (msg: string) => {
     core.debug(msg);
     core.info(msg);
   };
+
+  if (config.refresh) {
+    core.startGroup(`Refresh stack on ${config.stackName}`);
+    await stack.refresh({ onOutput });
+    core.endGroup();
+  }
+
+  core.startGroup(`pulumi ${config.command} on ${config.stackName}`);
 
   const actions: Record<Commands, () => Promise<string>> = {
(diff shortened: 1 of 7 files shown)


stack72 (commented):
This needs added to the README as well :)

src/main.ts
```diff
@@ -37,13 +37,19 @@ const main = async () => {
...
     : LocalWorkspace.selectStack(stackArgs));
 
-  core.startGroup(`pulumi ${config.command} on ${config.stackName}`);
-
   const onOutput = (msg: string) => {
     core.debug(msg);
     core.info(msg);
   };
 
+  if (config.refresh) {
```
stack72, line 45: I really like this - separating the refresh out of the log group :)

stack72 (approved):

src/main.ts
cobraz: Yeah, thanks!

PR #134, opened 2021-03-22
feat: Add improved configuration logging

cobraz:


src/main.ts
```diff
@@ -9,7 +9,7 @@ import { invariant } from './libs/utils';
 
 const main = async () => {
   const config = await makeConfig();
-  core.debug('Configuration is loaded');
+  core.debug(`Configuration is loaded: ${JSON.stringify(config, null, 2)}`);
```
stack72, line 12: Is there any chance this will leak any secrets?

cobraz: AFAIK, no. It should be masked automatically.

This debug wouldn't be echoed without the user explicitly [asking for it](https://docs.github.com/en/actions/managing-workflow-runs/enabling-debug-logging) either.

cobraz: Reference:
https://docs.github.com/en/actions/reference/encrypted-secrets

> Warning: GitHub automatically redacts secrets printed to the log, but you should avoid printing secrets to the log intentionally.

PR #154, opened 2021-03-31
ci: Improve workflow

cobraz:
The way our current workflow is configured, we hit the maximum concurrent builds all day. This pull request replaces all the separate workflows into one, where building our Github Action is required to succeed before testing on all the different stacks.

IMHO, it would be much better to have them separated as before, but AFAIK we are not able to reference jobs in other workflows.

Since testing on stacks requires a successful build, we get less waiting for builds to succeed (queued builds). If something fails in Install and Build GHA (`install-and-build`), the whole workflow will _fail_ (actually, the `install-and-build` job will fail, but the rest of the jobs will be skipped).

### Other features

- Added cache to NPM, Python and to the Install and Build GHA job (build time = seconds now!).
- Added a warning message if dist files are not changed (let's say a user contributes something, but it does not change stuff in `dist/`, that is something we should know)


stack72:
This is awesome :D

PR #181, opened 2021-04-15
refactor: Improve inputs

cobraz:


stack72 (approved):
Oh I like this! We should make sure and document that is the case though in our README

cobraz:
> Oh I like this! We should make sure and document that is the case though in our README

Yes :)

PR #211, opened 2021-05-04
feat: Install Pulumi CLI

### LINKED ISSUE #210: Embed Installation of the CLI as part of the Action
We have decided that embedding the installation of the CLI as part of this action is the best course of action to solve the versioning requirements of theaction.

The CLI version itself has no real bearing on the outcome of the user interaction - therefore we can ensure that we make sure the CLI version is always at the correct version before the action will execute the users Pulumi Program

cobraz:
With this pull request, you don't have to install Pulumi CLI. Also adds a fix to the workflow, which didn't catch all code changes.

- [x] Merge pulumi/docs#5923 (required to get `versions.json`)
- [x] Change the URL to `versions.json` after 👆

fixes #210


src/config.ts
```diff
@@ -31,6 +31,7 @@ export const config = rt
     workDir: rt.String,
     commentOnPr: rt.Boolean,
     options: options,
+    pulumiVersion: rt.String,
```
stack72, line 34: I don't think we even need to expose this - I think what we should do is internally set this to match 3.x and make sure we always install the new 3.x version

Allowing a user to control this may still cause problems for us with versions being mismatched

A CLI version will have no real bearing on the program itself

src/libs/pulumi-cli.ts
```diff
@@ -8,3 +14,46 @@ export async function isAvailable(): Promise<boolean> {
 export async function run(...args: string[]): Promise<void> {
   await exec.exec(`pulumi`, args, true);
 }
+
+export async function downloadCli(range: string): Promise<void> {
```
stack72, line 18: what happens if there is already a version installed - will this new install overwrite that?

src/config.ts
cobraz: This is not a required field, and it defaults to `^3` now.

We could make sure the version set by the user qualifies (the range that works with a particular version of the GHA). We probably don't need that if we're only doing this internally.

stack72: we are going to be doing this internally so that users don't care about this anymore

src/libs/pulumi-cli.ts
cobraz: AFAIK no, both will be installed, both in `PATH`. We could validate if it is available and exit, or potentially deterministically delete (expect it to be in `.pulumi`?)

cobraz: We could also fail if `pulumi` is installed?

stack72: So for this, we will ensure we use the version that the action installs internally

src/main.ts
```diff
@@ -7,12 +7,13 @@ import { handlePullRequestMessage } from './libs/pr';
...
 
+const pulumiVersion = '^3';
+
 const main = async () => {
   const config = await makeConfig();
   core.debug('Configuration is loaded');
 
-  invariant(pulumiCli.isAvailable(), 'Pulumi CLI is not available.');
-  core.debug('Pulumi CLI is available');
+  await pulumiCli.downloadCli(pulumiVersion);
```
stack72, line 17: should we add a debug message about what version is now available?

cobraz: We actually output the range in the `downloadCli` function, along with matched version.

stack72: wonderful!!

stack72 (approved):
we can squash and merge! I will follow up with some new test workflow and docs changes :)

PR #247, opened 2021-05-24
Bumping version of @pulumi/pulumi to 3.3.0

stack72:
Closes: #242
Closes: #244
Closes: #246
Fixes: #235

This allows us to set a userAgent to the call of pulumi/actions@v3
this is only used in understanding the usage of the action itself
not the user usage


README.md
```diff
@@ -20,29 +20,22 @@ jobs:
...
+This will check out the existing directory and run `pulumi preview`.
 
 ## Configuration
 
 The action can be configured with the following arguments:
 
-- `command` (required) - The command to pass to the Pulumi CLI. Accepted values
-  are `up` (update), `refresh`, `destroy` and `preview`. This command is the
-  equivalent of running `pulumi <command>` if your terminal.
+- `command` (required) - The command to run as part of the action. Accepted
```
cobraz, line 37: Nice catch!

src/config.ts
```diff
@@ -63,6 +63,7 @@ export async function makeConfig(): Promise<Config> {
       target: parseArray(getInput('target')),
       targetDependents: parseBoolean(getInput('target-dependents')),
       editCommentOnPr: parseBoolean(getInput('edit-pr-comment')),
+      userAgent: "pulumi/action@v3"
```
cobraz, line 66: Should probably add `userAgent` to the run-type in the same file (`options`). It's probably good enough with an `rt.Literal('pulumi/actions@v3')`.

stack72: Fixed and added the userAgent to the tests

cobraz (approved):

stack72:
These tests are failing due to the missing changes in dist which the PR doesn't allow

PR #250, opened 2021-05-24
No need to delete Pulumi homedirectory when installing CLI

stack72:
Pulumi will install over the top of what is already there and we
will "upgrade" the existing version

### DIFF
--- a/src/libs/pulumi-cli.ts
+++ b/src/libs/pulumi-cli.ts
@@ -1,6 +1,7 @@
 import * as os from 'os';
 import * as path from 'path';
 import * as core from '@actions/core';
+import * as io from '@actions/io';
 import * as tc from '@actions/tool-cache';
 import * as exec from './exec';
 import { getVersionObject } from './libs/get-version';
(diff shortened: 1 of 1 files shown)


src/libs/pulumi-cli.ts
```diff
@@ -1,8 +1,9 @@
+import * as fs from "fs";
 import * as os from 'os';
 import * as path from 'path';
 import * as core from '@actions/core';
-import * as io from '@actions/io';
 import * as tc from '@actions/tool-cache';
+import * as makeDir from 'make-dir';
```
cobraz, line 6: `@actions/io` supports `await io.mkdirP('path/to/make');` and can also do `await io.mv('path/to/file', 'path/to/dest');`.

src/libs/pulumi-cli.ts
```diff
@@ -37,22 +38,31 @@ export async function downloadCli(range: string): Promise<void> {
...
 
   const downloaded = await tc.downloadTool(downloads[platform]);
   core.debug(`successfully downloaded ${downloads[platform]}`);
 
-  const extractedPath = await tc.extractTar(downloaded, destination);
-  core.debug(`Successfully extracted ${downloaded} to ${extractedPath}`);
+  switch (platform) {
+    case "windows": {
+      await tc.extractZip(downloaded, os.homedir());
+      fs.renameSync(path.join(os.homedir(), "Pulumi"), path.join(os.homedir(), ".pulumi"));
```
cobraz, line 57:
```suggestion
      await io.mv(path.join(os.homedir(), "Pulumi"), path.join(os.homedir(), ".pulumi"));
```

src/libs/pulumi-cli.ts
```diff
@@ -1,8 +1,9 @@
+import * as fs from "fs";
```
cobraz, line 1:
```suggestion
```

src/libs/pulumi-cli.ts
```diff
@@ -37,22 +38,31 @@ export async function downloadCli(range: string): Promise<void> {
...
+      break;
+    }
+    default: {
+      const destinationPath = await makeDir(destination);
+      core.info(`Successfully created ${destinationPath}`)
+      const extractedPath = await tc.extractTar(downloaded, destination);
+      core.info(`Successfully extracted ${downloaded} to ${extractedPath}`)
+      const oldPath = path.join(destination, "pulumi")
+      const newPath = path.join(destination, "bin")
+      fs.renameSync(oldPath, newPath);
```
cobraz, line 67:
```suggestion
      await io.mv(oldPath, newPath);
```

src/libs/pulumi-cli.ts
```diff
@@ -37,22 +38,31 @@ export async function downloadCli(range: string): Promise<void> {
...
-  const extractedPath = await tc.extractTar(downloaded, destination);
-  core.debug(`Successfully extracted ${downloaded} to ${extractedPath}`);
+  switch (platform) {
+    case "windows": {
+      await tc.extractZip(downloaded, os.homedir());
+      fs.renameSync(path.join(os.homedir(), "Pulumi"), path.join(os.homedir(), ".pulumi"));
+      break;
+    }
+    default: {
+      const destinationPath = await makeDir(destination);
```
cobraz, line 61:
```suggestion
      const destinationPath = await io.mkdirP(destination);
      
```

src/libs/pulumi-cli.ts
```diff
@@ -36,15 +38,31 @@ export async function downloadCli(range: string): Promise<void> {
   const { version, downloads } = await getVersionObject(range);
   core.debug(`Matched version: ${version}`);
 
-  const destination = path.join(os.homedir(), '.pulumi');
-  core.debug(`Install destination is ${destination}`);
+  const destination = path.join(os.homedir(), ".pulumi");
```
cobraz, line 41:
```suggestion
  const destination = path.join(os.homedir(), '.pulumi');
```

cobraz (commented):
I generally like when we're using `@actions/*` packages, as they're made by Github to use with Github Actions. I added a few examples where we replace `fs` and `make-dir` with `@actions/io`.

I see there are a few double quotes, maybe you can format the document with Prettier before merging?

cobraz (approved):
🚀

PR #254, opened 2021-05-25
Bump pulumi/pulumi dependency

stack72:
Closes: #251
Closes: #252
Closes: #253


src/libs/__tests__/__snapshots__/get-version.test.ts.snap
```diff
@@ -1,16 +1,16 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`get-version range versions should match ^3 versions 1`] = `"v3.3.0"`;
+exports[`get-version range versions should match ^3 versions 1`] = `"v3.3.1"`;
```
cobraz, line 3: We should need to update the snaps for simple version bumps. 🤔

cobraz (approved):

PR #281, opened 2021-06-11
Add missing input

cobraz:
This pull request adds missing input (`edit-pr-comment`).


action.yml
```diff
@@ -59,6 +59,10 @@ inputs:
     description: 'Create the stack if it currently does not exist'
     required: false
     default: 'false'
+  edit-pr-comment:
+    description: 'Edit previous PR comment instead of posting new one'
+    required: false
+    default: 'false'
```
cobraz, line 65: We should probably set `edit-pr-comment` to default `true`. From a new users perspective, it's better the comment is updated than spamming the comments.

```suggestion
    default: 'true'
```

stack72 (approved):

action.yml
cobraz: @stack72 What do you think about making this default on?

PR #289, opened 2021-06-18
feat: print matched version

cobraz:
We should probably print which version that is running, so it's easier to debug.


stack72 (approved):

cobraz:
Tests was already failing before this, so merging this. I'll open a new PR on it soon :)

PR #496, opened 2022-01-21
Allow setting pr via with.comment-on-pr-number

hermanbanken:
[When reusing workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) one loses the event payload of the original Workflow, ergo, `payload.pull_request` might be undefined. The invariant would fail the action, but we can do better!

This allows users to set the PR number explicitly, so the Pulumi Action can be part of a larger workflow.


hermanbanken:
Having the dist files in the branch is the only way to easily test the action myself before submitting the PR. Won't the GitHub Action just overwrite the dist regardless?

src/libs/pr.ts
```diff
@@ -24,15 +24,16 @@ export async function handlePullRequestMessage(
   `;
 
   const { payload, repo } = context;
-  invariant(payload.pull_request, 'Missing pull request event data.');
+  const nr = config.commentOnPrNumber || payload.pull_request.number;
```
cobraz, line 27:
```suggestion
  const nr = config.commentOnPrNumber ?? payload.pull_request.number;
```

src/config.ts
```diff
@@ -55,6 +56,7 @@ export async function makeConfig(): Promise<Config> {
     cloudUrl: getInput('cloud-url'),
     githubToken: getInput('github-token'),
     commentOnPr: parseBoolean(getInput('comment-on-pr')),
+    commentOnPrNumber: parseNumber(getInput('comment-on-pr-number', {})),
```
cobraz, line 59: Why'd you need `{}` here?

action.yml
```diff
@@ -19,6 +19,10 @@ inputs:
     description: 'If true, a comment will be created with results'
     required: false
     default: 'false'
+  comment-on-pr-number:
+    description: 'Overrides the PR used to comment on'
+    required: false
+    default: '0'
```
cobraz, line 25: Do we need to set a default?

cobraz (commented):
I left some nitpicking, but it's probably fine as-is too. :)

cobraz:
Should we try to land this?
I would love to get the questions answered, just so I understand better what's going on.

Can you also fix the conflicts?

Thank you again!

gudleik:
would love to see this released, as we just bumped into the issue using reusable workflows.

hermanbanken:
Fixed & rebased!

action.yml
cobraz:
```suggestion
```

I think we probably could leave it out, but I can do that later (I'm going to refactor some of our config setup)

cobraz (approved):
I think this is ready now! It looks good, thank you very much!

RobbieMcKinstry:
Our current version of Node doesn't support the nullish coalescing operator, which I believe is why CI is failing.
https://github.com/pulumi/actions/actions/runs/3055194160/jobs/4927977989#step:6:115

RobbieMcKinstry:
I fixed the failing test and rebased this PR in https://github.com/pulumi/actions/pull/748

PR #562, opened 2022-04-05
chore: Add pr-number input

pellizzetti:
### What does this pull request do?
Add a new input called `pr-number` as an extra option.

When `comment-on-pr` is `true` it will check if the payload contains a pull request number. If it doesn't find, it will try to assign a value from `pr-number`. If both don't exist, it will fail with the current `Missing pull request event data.` message.

### Why?
I run a lot of workflows with `repository_dispatch` and I would like to receive comments on the PR that triggered the event.

### How has it been tested?
_Dispatcher_: https://github.com/pellizzetti/actions/blob/master/.github/workflows/pr_number.yml
_Receiver_: https://github.com/pellizzetti/actions/blob/master/.github/workflows/pr-number-event.yml

[PR](https://github.com/pellizzetti/actions/pull/2) with comment returned on the _Receiver_ workflow.


pellizzetti:
Closing this in favor of #496 😆

Issue #633, opened 2022-06-02
Pull Request Comments are not Replaced on Subsequent Runs


canhorn:
### What happened?

Pull Request comments are not replaced on subsequent runs.

When Summary/Details were implemented, they were added into the heading. https://github.com/pulumi/actions/blob/0c9ec5ff167fe048d2f54775bf7e19a559d6e133/src/libs/pr.ts#L18-L21

This probably causes the `comments.find` to never find the comment, and always create a new comment on the PR.

### Steps to reproduce

Trigger multiple Workflow run, with Comment on PR set to `true`.

```
jobs:
  preview_production:
    name: Preview Production Environment
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v2

      - uses: actions/setup-dotnet@v1
        with:
          dotnet-version: 6.0.x

      - uses: pulumi/actions@v3
        with:
          command: preview
          stack-name: production
          comment-on-pr: true
          diff: true
```

### Expected Behavior

The first comment about a run to be replaced on subsequent runs.

### Actual Behavior

Multiple Comments added to PR, one for every run.

### Versions used

Current

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

shreecodes:
This issue has made some of our infrastructure PRs very noisy.

I noticed that #635 was merged. Are you folks planning a new release soon? 😄

RobbieMcKinstry:
Hi @shreecodes! We're having an intern discussion about release cadence for this repo. I'm new to the team, so I need to come up to speed on the release process for this repo, but I hope we can cut a release soon since this seems to be an annoyance for a couple of people.

I'm going to close this issue, as it was fixed with #635, but I'll continue the conversation around cadence to #669.

PR #635, opened 2022-06-05
fix: update PR comment

jbrunton:
This is a fix for #633.

The problem appears to be from the usage of dedent, which produces mixed indentation levels between `heading` and `body`. On current master, one ends up with a `heading` that looks like this:

```
#### :tropical_drink: `up` on pr-comment

<details>
<summary>Click to expand Pulumi report</summary>
```

and a `body` that looks like this (note the extra indentation):
```
#### :tropical_drink: `up` on pr-comment
   <details>
   <summary>Click to expand Pulumi report</summary>
```

I considered amending the indentation of the `body` string (basically, removing it all), but that would require some unusual formatting that would break the flow of the code, and seems brittle. This change should be a little more robust.

Ideally there would be unit test coverage for this case, but I didn't have time to add it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,15 @@
 
 ## HEAD (Unreleased)
 
-_(none)_
+- fix: update PR comments correctly when `edit-pr-comment` is true (fixes
+  [#633](https://github.com/pulumi/actions/issues/633))
 
 ---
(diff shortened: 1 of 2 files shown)


cobraz (approved):
Thanks for this!

PR #648, opened 2022-06-11
chore: use upsert for all steps

cobraz:


cobraz:
Since this change didn't trigger Actions Workflow, I've manually triggered them here: https://github.com/pulumi/actions/actions/runs/2480060986

cobraz:
> Since this change didn't trigger Actions Workflow, I've manually triggered them here: [pulumi/actions/actions/runs/2480060986](https://github.com/pulumi/actions/actions/runs/2480060986)

That obviously didn't work, so I created #649 to address that.

stack72 (approved):

PR #651, opened 2022-06-11
chore: bump to node@16

cobraz:
Unsure if this needs to be a breaking change, but probably smart?

### DIFF
--- a/.github/workflows/update_dist.yml
+++ b/.github/workflows/update_dist.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 16.x
           cache: yarn
       - run: yarn install
       - run: yarn build
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -19,7 +19,7 @@ jobs:
 
       - uses: actions/setup-node@v3
         with:
-          node-version: 14.x
+          node-version: 16.x
           cache: yarn
 
       - run: yarn install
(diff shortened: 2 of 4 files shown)


t0yv0 (approved):
LGTM, thank you. It seems like it's time to update. I found https://endoflife.date/nodejs which is helpful.

stanislav-zaprudskiy:
With this change GH Enterprise versions before 3.4 brake - `node16` isn't supported there. https://docs.github.com/en/enterprise-server@3.3/actions/creating-actions/metadata-syntax-for-github-actions#runsusing

> Error: System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter ''using: node16' is not supported, use 'docker' or 'node12' instead.')
   at GitHub.Runner.Worker.ActionManifestManager.ConvertRuns(IExecutionContext executionContext, TemplateContext templateContext, TemplateToken inputsToken, String fileRelativePath, MappingToken outputs)
   at GitHub.Runner.Worker.ActionManifestManager.Load(IExecutionContext executionContext, String manifestFile)
Error: Fail to load ghcom-actions/pulumi-actions/v3/action.yml

t0yv0:
Node 12 support ended 30 Apr 2022

Would it be possible to upgrade to enterprise-server @3.5 that supports node 16?

https://docs.github.com/en/enterprise-server@3.5/actions/creating-actions/metadata-syntax-for-github-actions#runsusing

t0yv0:
Talked to Luke, we might be open to support node a few months after EOL. Opened a PR to revert for now.

stanislav-zaprudskiy:
In our case GHE upgrade is not something which I could do or request - it's operated by different team. But GHE versions also have EOL timelines - https://docs.github.com/en/enterprise-server@3.5/admin/all-releases#currently-supported-releases. Even `3.3` doesn't support `node16`, and according to the doc it's supposed to remain non-deprecated until `2022-12-07`.

Anyway, thank you for a fast fix for this!

PR #658, opened 2022-06-16
feat: add support for local policy packs

### LINKED ISSUE #212: Allow usage of local pulumi policy pack
We are trying to move from running plain pulumi cli in github actions to the actual `pulumi/action` since that would support editing/replacing the same PR comment instead of flooding the PR comments by the pulumi github app.

But the one feature I seem to be missing is the equivalent of the `--policy-pack ../pulumi-policy` CLI argument. We run a (large) monorepo and have the policy files in the same repo and refer to the locally when doing a pulumi up or preview.

jaxxstorm:
This adds support for specifying local policy packs and policy pack configs.

It needs to ship _after_ https://github.com/pulumi/pulumi/pull/9872 ships into a new version with the automation API deps bumped

Fixes #212


cobraz:
> @cobraz do you know why the OS X tests are failing?

We need #648 to be landed. Can you do a review on that maybe?

cobraz:
We should probably change the pulumi version required by the actions before releasing.

https://github.com/pulumi/actions/blob/ea444edc9d0c81cc7883c95bed84b67829c67503/src/main.ts#L14

Maybe something like `^3.35` would suffice? Just so people don't start using it with older versions? This will of course be more relevant if/when #661 is landed.

cobraz:
@jaxxstorm Can you rebase, hopefully that will fix the builds :)

Moon1706:
@jaxxstorm  Hi!
As I know, additional functionality really slooooowly appended to this Action. I try to add configMap 2 month. So, if it's critical for you, please, use [my fork](https://github.com/terraform-gitlab-modules/actions) to add new features. I hope in the future it will be an append to the official action.

PR #678, opened 2022-07-07
Revert node16 upgrade

t0yv0:
Based on discussion in https://github.com/pulumi/actions/pull/651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## HEAD (Unreleased)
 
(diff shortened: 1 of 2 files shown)


RobbieMcKinstry (approved):
Ah, I see we used Node 14 in a couple of places, and bumped those to Node 16. Those locations are strictly for our own workflows (the workflows for this repository), yes? And thus all we actually need to revert is the Node version used in the client workflows.

Assuming that's true, LGTM.

Issue #677, opened 2022-07-07
3.18.0 fails when not run on pull request and comment-on-pr is true


ppalucha:
### What happened?

Starting with version 3.18.0 if 'comment-on-pr' is set to true and action is triggered on non-pr event, it fails (I edited out actual paths):
```
pulumi up on prod
.../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/node_modules/ts-invariant/lib/invariant.cjs:16
        var _this = _super.call(this, typeof message === "number"
^
Invariant Violation: Missing pull request event data.
    at new InvariantError (.../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/node_modules/ts-invariant/lib/invariant.cjs:16:1)
    at invariant (.../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/node_modules/ts-invariant/lib/invariant.cjs:28:1)
    at .../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/src/libs/pr.ts:42:1
    at Generator.next (<anonymous>)
    at .../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/node_modules/tslib/tslib.js:117:1
    at new Promise (<anonymous>)
    at __awaiter (.../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/node_modules/tslib/tslib.js:113:1)
    at handlePullRequestMessage (.../_work/_actions/pulumi/actions/v3/dist/index.js:88219:12)
    at .../_work/_actions/pulumi/actions/v3/webpack:/pulumi-github-action/src/main.ts:97:1
    at Generator.next (<anonymous>)
```

### Steps to reproduce

Configure workflow e.g. as:
```
on:
  push:
    branches:
      - main
```

Configure action as:
```
 - uses: pulumi/actions@v3
    with:
      command: up
      stack-name: prod
      work-dir: workdir
      cloud-url: s3://my-bucket/my-folder
      refresh: true
      comment-on-pr: true
```

### Expected Behavior

Up command running without errors.

### Actual Behavior

Error as in 'What happened' section (`Invariant Violation: Missing pull request event data.')

### Versions used

3.18.0

### Additional context

It used to work on 3.18.0; removing `comment-on-pr` fixes the issue.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

[...]

PR #679, opened 2022-07-07
Only "Comment-On-PRs" when running in a PR context

RobbieMcKinstry:
v3.18.0 introduced a bug where Actions that supplied `comment-on-true: true` would panic when not running within the context of a PR (_i.e._ `event type == pull_request`). This resulted in panics for users with `pulumi@v3` in their workflows, which is fairly common as it's what the docs suggest to do.

This PR adds a guard to check if we're in a PR context before issuing comments.

Fixes https://github.com/pulumi/actions/issues/677 which is a P1.


RobbieMcKinstry:
Ahhh welp, this change didn' fix my repro... that's very sad.
The PR is missing a regression test but God darn I really thought this was the root of the issue.

AaronFriel (approved):

Issue #707, opened 2022-08-12
Document Version Support for Dependent Tools


RobbieMcKinstry:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

This Action [runs on Node12](https://github.com/pulumi/actions/blob/master/action.yml#L91-L99), which is [currently EOL](https://github.com/nodejs/Release/#end-of-life-releases). GitHub Actions supports only Node 12 and Node 16 ([Node 14 is not supported](https://docs.github.com/en/actions/creating-actions/metadata-syntax-for-github-actions#runsusing)).

Some of our customers use GitHub Enterprise, which has its own [maintenance schedule](https://docs.github.com/en/enterprise-server@3.5/admin/all-releases#currently-supported-releases). We want to make sure users who are on all supported versions of GitHub Enterprise can use the action, even if the Node runtime is EOL. GitHub Enterprise [added support for Node16](https://docs.github.com/enterprise-server@3.4/admin/release-notes#github-actions-changes) in v3.4.0. Some of our customers use GitHub Enterprise 3.3 and 3.2, which are both under active maintenance.

This issue is to improve our documentation for GitHub Enterprise support, it's interactions with Node, and to document our process around upgrading.

_n.b.:_ We don't need to publicly document our usage of Yarn or TypeScript since those are build-time dependencies. (TIL GH Actions only allow pure JavaScript and invokes `dist/index.js` directly.) 

### Affected area/feature

* External Documentatin: namely, update the README to indicate our policy around NodeJS support.

* Internal Documentation: We should schedule a reminder to upgrade NodeJS, and establish an short playbook about the maintenance process.

PR #711, opened 2022-08-21
Document sunsetting policy for Node 12

RobbieMcKinstry:
This PR documents our intention to support our enterprise customers. As part of a "Pulumi EOL" issue, I did some reading to determine what kind of policies we can publish around version support -- in the case of this repo, only Node support is required to disclose, since yarn, typescript, etc are all build-time dependencies.

It also provides a public announcement of our intention to end support for Node 12 in December.

Closes https://github.com/pulumi/actions/issues/707

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,3 +224,12 @@ As of `v3.18`, we are intending to move to a monthly cadence for minor releases.
 Minor releases will be published around the beginning of the month. We may cut a
 patch release instead, if the changes are small enough not to warrant a minor
 release. We will also cut patch releases periodically as needed to address bugs.
+
+## Node 12 Support Ending 2022-12-07
+
+We intend this action to work for all GitHub users, both SaaS users and GitHub
+Enterprise users. As such, this action will continue to use Node 12 until all
+maintained versions of GitHub Enterprise have Node 16 support. This date is
+2022-12-07, when GitHub Enterprise v3.3 is sunset. After 2022-12-07, this action
+will upgrade to Node 16, and will no longer run on GitHub Actions runners
+without Node 16 support.


cobraz:
Could we do a major bump and start running node v16 today?

RobbieMcKinstry:
Hmm, that's a good point! I think that's preferred over continuing to support Node 12, since it has been sunset.

RobbieMcKinstry:
I asked around internally, and here's a summary of the conversation:

One person expressed concern that users who don't specify a version of the action will get automatically upgraded to `v4` and will still experience the breaking change. I feel that comes with the territory of not pinning a major version...

Someone else suggested we cut a release `@v3-nodejs12` and tell customers to pin to that if they need Node 12, then continue to release on the `v3` line. That requires coordination with the customers since we'd have to communicate to our CX engineers that users need to pin to that version. Then, we'd have to give them some time to implement the change before we cut the new release. I presume it would also be preferable for us to communicate a v4 release before it happens. 🤷🏻‍♂️

I noted that if we're going to be blocked until users make the pinning change, we might as well wait until December anyway, at which point GH Enterprise v3.3 will have been deprecated, and we can upgrade without issue. Those in the discussion agreed that we should broadcast the expected change now (via a pinned GH Issue and documentation in the README). So the conclusion was essentially to merge this PR and pin an issue to the top of the backlog to prepare users for the migration. 😕

t0yv0:
If this helps, for context https://github.com/pulumi/actions/pull/651 - we can request feedback from Stanislav if we go that route. There might be more users affected but that's the one user we definitely heard from before when upgrading to 16.

t0yv0 (approved):
Decided to accept as is for the moment. Please let me know if any features are blocked on the upgrade.

Issue #698, opened 2022-08-01
Error - "PULUMI_ACCESS_TOKEN must be set" even if cloud-url param was used


pr-a4bee:
### What happened?

I'm trying to add infrastructure check with PR comment during CI pipeline. For that, Pulumi/actions@v3 was used. Unfortunately, it returns errors:

> stderr: error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions

### Steps to reproduce

According to documentation, action with cloud-url was used. AWS credentials were set directly as env and with aws-actions/configure-aws-credentials@v1 - with same outcome.

`- uses: pulumi/actions@v3.18.1
     env:
          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
          AWS_REGION: ${{ secrets.AWS_REGION }}
          PULUMI_CONFIG_PASSPHRASE: ''
        with:
          command: preview
          stack-name: ${{ env.PULUMI_STACK }}
          comment-on-pr: true
          cloud-url: ${{ env.PULUMI_BUCKET }}
          pulumi-version: ${{ env.PULUMI_VERSION }}
          work-dir: infrastructure/backend
`

### Expected Behavior

Pulumi is showing infrastructure changes during PR

### Actual Behavior

Pipeline fails

### Versions used

v3

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

mikhailshilkov:
Hi @pr-a4bee, thank you for opening an issue!

To be sure - what is the value that you are trying to set `cloud-url` to? Are you trying to login to an S3 bucket?

pr-a4bee:
Yes, value was set to: s3://xxx-xxx. It works fine with custom pulumi login script.

RobbieMcKinstry:
Hi @pr-a4bee I went ahead and attempted to reproduce this.
I was able to reproduce the error message by introducing a type in the line 

```yml
   cloud-url: ${{ env.PULUMI_BUCKET }}
```

When `PULUMI_BUCKET` is set to an empty string, I get the error message `PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions`. Or, when `PULUMI_BUCKET` references an environment variable that doesn't exist, I get the same error.

[...]

PR #717, opened 2022-09-01
fix logins for self-hosted Pulumi backend

tmehlinger:
Fixes https://github.com/pulumi/actions/issues/698.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,8 @@ _(none)_
 
 ## 3.19.1 (2022-09-01)
 
-- bug: Support logging into self-hosted Pulumi (i.e. Business Critical)
+- bug: Support logging into self-hosted Pulumi (e.g. an S3 bucket or
+  Pulumi Business Critical self-hosted)
   [#717](https://github.com/pulumi/actions/pull/717)
 
 ## 3.19.0 (2022-08-10)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,7 @@ _(none)_
 
 ## 3.19.1 (2022-09-01)
 
-- bug: Support logging into self-hosted Pulumi (e.g. an S3 bucket or
-  Pulumi Business Critical self-hosted)
+- bug: Fix bug where `cloud-url` option was not respected. This change also enables using self-hosted Pulumi (_e.g._ an S3 bucket or Pulumi Business Critical self-hosted)
   [#717](https://github.com/pulumi/actions/pull/717)
 
 ## 3.19.0 (2022-08-10)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ _(none)_
 
 ---
 
+## 3.19.1 (2022-09-01)
+
+- bug: Support logging into self-hosted Pulumi (i.e. Business Critical)
+  [#717](https://github.com/pulumi/actions/pull/717)
+
 ## 3.19.0 (2022-08-10)
 
 - feat: add configuration of Pulumi stack by `config-map` field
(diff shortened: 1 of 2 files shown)


src/main.ts
```diff
@@ -19,12 +19,12 @@ const main = async () => {
...
-  if (environmentVariables.PULUMI_ACCESS_TOKEN !== '') {
-    core.debug(`Logging into Pulumi`);
-    await pulumiCli.run('login');
-  } else if (config.cloudUrl) {
+  if (config.cloudUrl) {
     core.debug(`Logging into ${config.cloudUrl}`);
     await pulumiCli.run('login', config.cloudUrl);
+  } else {
+    core.debug(`Logging into Pulumi`);
+    await pulumiCli.run('login');
```
RobbieMcKinstry, line 27: I think we still need the first conditional guard as an `else if` branch to preserve the original behavior, no?

RobbieMcKinstry (changes requested):

RobbieMcKinstry:
Hi @tmehlinger Thanks very much for your submission!

src/main.ts
tmehlinger: `else if` to check whether `PULUMI_ACCESS_TOKEN` is set? If so... I think you're right, because the `pulumi login` command would block indefinitely on the prompt for a token if you were trying to login to a Pulumi backend (either cloud- or self-hosted). I can work on a fix in the morning.

jre21: On that note, are there any scenarios under which the remainder of the script can do any useful work if `pulumi login` hasn't been called?  If not, I'd think we'd want to throw an exception in an `else` block rather than wait for later commands to fail.

jre21: Although thinking on this a little more, the "login isn't necessary" case probably encompasses all the non-Pulumi-enterprise backends, in which case we definitely want to preserve the existing behavior of just continuing without logging in if neither condition matches.

CHANGELOG.md
```diff
@@ -6,6 +6,12 @@ _(none)_
 
 ---
 
+## 3.19.1 (2022-09-01)
+
+- bug: Support logging into self-hosted Pulumi (e.g. an S3 bucket or
+  Pulumi Business Critical self-hosted)
+  [#717](https://github.com/pulumi/actions/pull/717)
```
RobbieMcKinstry, line 13:
```suggestion
- bug: Fix bug where `cloud-url` option was not respected. This change also enables using self-hosted Pulumi (_e.g._ an S3 bucket or Pulumi Business Critical self-hosted)
  [#717](https://github.com/pulumi/actions/pull/717)
```

src/main.ts
```diff
@@ -15,16 +15,23 @@ import * as pulumiCli from './libs/pulumi-cli';
...
 
   await pulumiCli.downloadCli(config.options.pulumiVersion);
 
-  if (environmentVariables.PULUMI_ACCESS_TOKEN !== '') {
-    core.debug(`Logging into Pulumi`);
-    await pulumiCli.run('login');
-  } else if (config.cloudUrl) {
+  if (usingPulumiBackend && environmentVariables.PULUMI_ACCESS_TOKEN == '') {
+    throw new Error('PULUMI_ACCESS_TOKEN must be set to login to the Pulumi backend');
+  }
```
RobbieMcKinstry, line 27: This looks correct since any Url starting with `https` _should_ be a Pulumi service. However, it's _technically possible_ to login without providing PULUMI_ACCESS_TOKEN, either by passing the access token via stdin, or by providing a `credentials.json` file in `$PULUMI_HOME`. While it's probably not going to break other users, I'd prefer to not introduce an exception here.

I think it's possible for `login` to succeed if you don't

RobbieMcKinstry:
Hello again @tmehlinger, if it's okay with you, I opened https://github.com/pulumi/actions/pull/719 to modify this PR. I added a test and removed the exception.

RobbieMcKinstry:
I will close this issue since https://github.com/pulumi/actions/pull/719 has been merged.

PR #719, opened 2022-09-02
Use cloud-url if provided, fallback to access token.

RobbieMcKinstry:


RobbieMcKinstry:
Tests are failing due to a fixed P0 on v3.39.0. Once v3.39.1 rolls out, I'll rerun the workflows.

src/login.ts
```diff
@@ -0,0 +1,12 @@
...
+import * as pulumiCli from './libs/pulumi-cli';
+
+export const login = async (cloudUrl: string, accessToken: string) => {  
+  if (cloudUrl) {
+    core.debug(`Logging into ${cloudUrl}`);
+    await pulumiCli.run('login', cloudUrl);
+  } else if (accessToken == '') {
+    core.debug(`Logging into Pulumi`);
+    await pulumiCli.run('login');
+  }
```
tmehlinger, line 11: If `cloudUrl` is not set and `accessToken` is non-empty, this function falls through and does nothing.

tmehlinger (changes requested):
I don't think this solves the problem. The presence of the access token isn't relevant for making any decisions about logging in, it only matters if it's required to log into a Pulumi (cloud or self-hosted) backend. I don't mind avoiding raising an exception if the environment variable isn't set because there are other ways to authenticate (like populating PULUMI_HOME with a volume containing credentials.json) but there needs to be _some_ logic to provide useful feedback to the user. One of the complaints in #698 is that the output was misleading and this doesn't address the absence of good feedback in an error condition.

src/login.ts
jre21: This code will be more maintainable and easier to reason about if we

1. keep the special casing to a minimum
2. let the login command do the error handling.

To that end, I would recommend changing the `else if` block to an `else` and adding `--non-interactive` to all the login commands.  That way, we can guarantee that `pulumi login` will be called and that if it doesn't have the information needed to complete the login it will abort immediately (presumably with a useful error message) instead of waiting for input that Actions can't provide.

tmehlinger:
Also worth noting (and I'm getting off in the weeds a bit here), this could be simplified substantially by fixing the Pulumi CLI to simply delete [this condition](https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/login.go#L124) and set the default `cloud-url` value here to `https://app.pulumi.com`. The linked code only raises an error; the [underlying logic](https://github.com/pulumi/pulumi/blob/d4e361b5bdd15c5b1131263c9df91950bcc2d050/pkg/backend/httpstate/backend.go#L81) ultimately returns a hard-coded default URL without traversing any special cases and the login behavior is identical for both the managed and self-hosted backends.

tl;d, this and the CLI could both be simpler. 😄

src/login.ts
RobbieMcKinstry: Looking into it, [you were right previously](https://github.com/pulumi/actions/pull/717/files#r961207899). There are scenarios where the Pulumi CLI does not need to login before performing operations. 

For example, as long as a stack has been selected, this `Pulumi.yaml` file works without logging in:

```yml
name: test-tracing
runtime: nodejs
description: A minimal TypeScript Pulumi program
backend:
  url: file://./data
```
...because the backend is a local file.

As a result, I think we should skip the login step if `PULUMI_ACCESS_TOKEN` and `cloud-url` are both empty.

RobbieMcKinstry: > If cloudUrl is not set and accessToken is non-empty, this function falls through and does nothing.

You're right, I'm missing a `!` here. We want to login when the token is set, I believe.

RobbieMcKinstry:
> but there needs to be some logic to provide useful feedback to the user

Hi @tmehlinger I added the `--non-interactive` flag as @jre21 suggested, which I hope improves the error message. When I run `pulumi` on the CLI with the `--non-interactive` flag and no local backend settings, if I'm logged out, I get...

```bash
$ pulumi --non-interactive preview
error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions
```

The command works if I run the same, but provide a file backend, even without running `pulumi login`.

Does that address your concern?

RobbieMcKinstry:
> Also worth noting (and I'm getting off in the weeds a bit here), this could be simplified substantially by fixing the Pulumi CLI to simply delete [this condition](https://github.com/pulumi/pulumi/blob/master/pkg/cmd/pulumi/login.go#L124) and set the default cloud-url value here in the action to https://app.pulumi.com. The linked code only raises an error; the [underlying logic](https://github.com/pulumi/pulumi/blob/d4e361b5bdd15c5b1131263c9df91950bcc2d050/pkg/backend/httpstate/backend.go#L81) ultimately returns a hard-coded default URL without traversing any special cases and the login behavior is identical for both the managed and self-hosted backends.

You're definitely right about that. That code is gnarley!

tmehlinger (commented):
I think this will do! :shipit: 🚀

Zaid-Ajaj (approved):
Looks great!

cobraz (approved):
🚀

tmehlinger (approved):

PR #755, opened 2022-10-15
feat: Add optional project_name to action

### LINKED ISSUE #563: Support multiple editable PR comments: Bug for Monrepos and Matrix Stategy actions
## Hello!

- To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already)

## Issue details

On Github Actions, if we use a matrix strategy to run a preview on PR, with `comment-on-pr = true` (and `edit-pr-comment = true`, the default value), for each check, the comment will override the same comment, resulting in at the end of the workflow, we have only one preview result: the last finished run.

### Steps to reproduce

On preview workflow:
```yaml
  preview:
    name: Pulumi - Preview
    runs-on: ubuntu-latest
    strategy:
      matrix:
        app: ["app1", "app2"]
    steps:
[...]
    - uses: pulumi/actions@v3
      with:
        command: preview
        comment-on-pr: true
        work-dir: ${{ matrix.app }}
[...]
```

Expected: 

Have 1 comment on PR, for each matrix value (here, 2)

Actual: 

The same comment is updated for each matrix value
image

Workaround:

Set `edit-pr-comment` to `false`, but the PR will be commented for each matrix value and each run of the pipeline... it can be verbose 

```yaml
  preview:
    name: Pulumi - Preview
    runs-on: ubuntu-latest
    strategy:
      matrix:
        app: ["app1", "app2"]
    steps:
[...]
    - uses: pulumi/actions@v3
      with:
        command: preview
        comment-on-pr: true
        edit-pr-comment: false
        work-dir: ${{ matrix.app }}
[...]
```

miguel-botelho:
This PR adds a new optional property called `project_name`, so that the comments created by the Github action are unique, based on current `stack_infra` + `project_name`. This will facilitate working with Pulumi actions inside monorepos, so that multiple jobs are ran in a PR and each job will have its own comment, given they are for different stacks or projects.
Besides that, also added a bit more tests around the pr spec.

On another note, I went with creating `project_name` as optional, not sure on what is the desired approach for this one. Also, if there's any way to find the name of the Pulumi project without the user specifying on the YML, I believe would be less intrusive (but I couldn't find by reading the project's source code a way to do so). Feedback is more than welcomed!

Fixes #563


src/libs/pr.ts
```diff
@@ -12,10 +12,14 @@ export async function handlePullRequestMessage(
...
     stackName,
+    projectName,
     options: { editCommentOnPr },
   } = config;
 
-  const heading = `#### :tropical_drink: \`${command}\` on ${stackName}`;
+  let heading = `#### :tropical_drink: \`${command}\` on ${stackName}`;
+  if (projectName) {
+    heading = `${heading} for project ${projectName}`
+  }
```
RobbieMcKinstry, line 22: LGTM. 

Nit: Assume `projectName` is set. Do you think the message would be clear enough if written as 

```markdown
#### :tropical_drink: \`${command}\` on ${projectName}/${stackName}`
```
...or do you think having `for project ${projectName}` is necessary for clarity?

I'm leaning towards "not clear enough" but WDYT?

RobbieMcKinstry (changes requested):
Hey, @miguel-botelho thank you very much for your contribution and especially for implementing this so quickly! 

> if there's any way to find the name of the Pulumi project without the user specifying on the YML, I believe would be less intrusive

IMO, I think we can get the project name programmatically using the Automation API. The [Project Settings](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/automation/#ProjectSettings) have a field for the project name. 

For example, we import items from the Automation API [here](https://github.com/pulumi/actions/blob/master/src/main.ts#L3-L8).

You should be able to get an instance of `ProjectSettings` using [this method](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/automation/#LocalWorkspace-projectSettings). I think this method is callable whenever, but you might want to wait until [after this line](https://github.com/pulumi/actions/blob/master/src/main.ts#L40-L42) so the stack is selected. Forgive me, I'm not very well-versed on the Automation API.

RobbieMcKinstry:
Oh, I also wanted to say thanks for fixing up the tests too!

src/libs/pr.ts
miguel-botelho: I think projectName/stackName is enough IMO, if the name of the project is something indicative. Not sure what's the "norm" for Pulumi project names, but for the ones I use, the simpler output is better 😄

CHANGELOG.md
```diff
@@ -9,6 +9,8 @@
 
 - feat: Always show login information
 
+- feat: Add project name to avoid having unintended edited comments
```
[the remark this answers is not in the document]
miguel-botelho: Thanks for the suggestion!

miguel-botelho:
> Hey, @miguel-botelho thank you very much for your contribution and especially for implementing this so quickly!
> 
> > if there's any way to find the name of the Pulumi project without the user specifying on the YML, I believe would be less intrusive
> 
> IMO, I think we can get the project name programmatically using the Automation API. The [Project Settings](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/automation/#ProjectSettings) have a field for the project name.
> 
> For example, we import items from the Automation API [here](https://github.com/pulumi/actions/blob/master/src/main.ts#L3-L8).
> 
> You should be able to get an instance of `ProjectSettings` using [this method](https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/pulumi/automation/#LocalWorkspace-projectSettings). I think this method is callable whenever, but you might want to wait until [after this line](https://github.com/pulumi/actions/blob/master/src/main.ts#L40-L42) so the stack is selected. Forgive me, I'm not very well-versed on the Automation API.

Hey @RobbieMcKinstry , many thanks for the help and the quick review. I ended up passing the `projectName` as a property of the current `Config` object, not sure if there's a better/cleaner approach, so feel free to leave your feedback on this one: https://github.com/pulumi/actions/pull/755/commits/905033c83c161c244e8f241489f19de036111ee3

miguel-botelho:
Another question, how can I test this development on one of my projects?

src/main.ts
```diff
@@ -41,6 +41,8 @@ const main = async () => {
     ? LocalWorkspace.createOrSelectStack(stackArgs, stackOpts)
     : LocalWorkspace.selectStack(stackArgs, stackOpts));
 
+  config.projectName = (await stack.workspace.projectSettings()).name;
```
cobraz, line 44: Maybe we should split this into two variables?

```suggestion
	const projectSettings = await stack.workspace.projectSettings();
  config.projectName = projectSettings.name;
```

Something like that? Just to improve readability?

Maybe we should add an invariant as well? I'm not sure if we do any validation in `projectSettings()`, so might not be needed.

cobraz (approved):
I think this looks good!

Thanks for this, I'm sure people will love this addition! Especially the ones with mono-repos with multiple stacks :)

src/config.ts
```diff
@@ -34,6 +34,7 @@ export const config = rt
     command: command,
     stackName: rt.String,
     workDir: rt.String,
+    projectName: rt.String,
```
RobbieMcKinstry, line 37: I think we can eliminate the `projectName` input entirely by relying on this line below:

```typescript
const projectSettings = await stack.workspace.projectSettings();
```

We should always be able to get the project name from the config.

RobbieMcKinstry: Ah, did you push that change? Right now, in the diff, I see the value in `Config` is always set to the empty string.
My goal is to eliminate partially initialized states in the `Config` object. As much as possible, that `Config` object should be immutable:
1. If at the time of construction, we can access the `projectName`, then that works.
2. But if not, I'd prefer passing the `projectName` field into the handlePullRequest` function.

From the diff I'm seeing, the config object is initialized with an empty string [here](https://github.com/pulumi/actions/pull/755/files#diff-c3095d5010e65c52737a98a5d618ea24049ebe90c8470752426081d70ed6e012R63) and that value is overwritten [here](https://github.com/pulumi/actions/pull/755/files#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dR44-R46). I'd prefer if we could avoid this mutation.

Does that help? 😅 Sorry if that's unclear or inconsistent.

RobbieMcKinstry:
> Another question, how can I test this development on one of my projects?

I think you can refer to the public URL and commit SHA (or branch name). 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-versioned-actions

src/config.ts
miguel-botelho: Got it! I think it's easier to pass projectName as a field on the function as well (and also avoids having a variable with a different value, depending on time of calling). I'll look into it in a bit, commit the changes, and test it as well via the link you sent. I'll notify in here when I've tested it "for real" 👍

miguel-botelho:
Hey @cobraz @RobbieMcKinstry 👋 
Many thanks for the review and help getting this one figured out. I've made the necessary changes to retrieve the `projectName` automatically and pass it down to the `handlePullRequest` method. I also went ahead and compiled the project in this branch (I've already reverted it) to test it on one of our monorepos, and it looks like it's working:

image

Feel free to review it again and leave your feedback on this one 😄

miguel-botelho:
Hey @RobbieMcKinstry , sorry for pinging you again, were you able to review this one? 🙏

miguel-botelho:
Hey @RobbieMcKinstry , just rebased it with the most recent changes in `master`, it's good to take a look again 👍

RobbieMcKinstry (approved):
Thank you for this contribution! I appreciate your thoughtful effort in working through my requested changes and your patience and persistence after I went dark on review. Seriously, thank you @miguel-botelho! 🙇🏻

RobbieMcKinstry:
@miguel-botelho Would you be willing to send me an email at robbie@pulumi.com? I want to send a small thank-you gift from Pulumi for your contribution.

miguel-botelho:
> @miguel-botelho Would you be willing to send me an email at [robbie@pulumi.com](mailto:robbie@pulumi.com)? I want to send a small thank-you gift from Pulumi for your contribution.

Many thanks! I'll ping you there in a few minutes. Also, I'm the one that should be thanking for guiding me during the review phase 🙇

RobbieMcKinstry:
Hah, anytime, friend!

PR #756, opened 2022-10-16
clarify stack name

susanev:
pairing this with https://github.com/pulumi/pulumi-hugo/pull/2054
super open to feedback if yall would prefer this included in a different way


cobraz (approved):
I think this makes sense! We also might wanna do the same for the examples, e.g. examples/python-pulumi.yaml?

PR #809, opened 2022-12-23
feat: added multi-comment-support on the same PR

philprime:
This adds a new configuration value to define a hidden comment identifier for pull-request comments.

The use-case for this feature showcased using this example:

```yaml
- uses: pulumi/actions@v3
  name: Preview the shared stack used by all environments
  with:
    command: preview
    stack-name: org-name/shared
    comment-on-pr: true
- uses: pulumi/actions@v3
  name: Preview the stack for the development environment
  with:
    command: preview
    stack-name: org-name/develop
    comment-on-pr: true
```

This workflow currently runs two preview commands but writes the latest output into the comment in the pull-request. By introducing a hidden HTML tag with an unique identifier, the workflow creates multiple comments:

```yaml
- uses: pulumi/actions@v3
  name: Preview the shared stack used by all environments
  with:
    command: preview
    stack-name: org-name/shared
    comment-on-pr: true
    comment-on-pr-id: hidden-shared-stack-id
- uses: pulumi/actions@v3
  name: Preview the stack for the development environment
  with:
    command: preview
    stack-name: org-name/develop
    comment-on-pr: true
    comment-on-pr-id: hidden-develop-stack-id
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## HEAD (Unreleased)
 
+- feat: Add multi-comment support on the same PR using comment identifiers
+  [#809](https://github.com/pulumi/actions/pull/809)
+
 --
 
 ## 4.3.0 (2023-05-12)
--- a/action.yml
+++ b/action.yml
@@ -23,6 +23,9 @@ inputs:
     description: 'If true, a comment will be created with results'
     required: false
     default: 'false'
+  comment-on-pr-id:
+    description: 'Hidden id to support multiple comments in one PR'
+    required: false
   comment-on-pr-number:
     description: 'Overrides the PR used to comment on'
     required: false
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -38,6 +38,7 @@ describe('config.ts', () => {
         "cloudUrl": "file://~",
         "command": "up",
         "commentOnPr": false,
+        "commentOnPrId": "",
         "commentOnPrNumber": undefined,
         "configMap": undefined,
         "editCommentOnPr": false,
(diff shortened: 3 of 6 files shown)


philprime:
Any feedback on this PR?

philprime:
We have been using our fork internally for 5 months now, creating comments for two stacks at the same time. I just rebased the fork on the latest upstream master. Any interest in getting this merged?

simenandre:
Hello, @philprime!

Thank you for opening this; sorry for not looking at it before now. Thanks so much for rebasing as well!

After #755 was landed and released in [v3.20.0](https://github.com/pulumi/actions/releases/tag/v3.20.0), we added the project and stack name to the comment heading, which should create one comment per stack as suggested by this pull request since the heading compares them.

Are there any other use cases I'm not seeing?

philprime:
Hi @simenandre,

from a quick glance at #755 I believe this could achieve the same goal, while my solution approach gives a more granular access, because the `id` could be any string, rather than only the stack itself. But I would agree there might not be a use case with two comments using the same stack.

I'll checkout the latest release and test it for our needs. Maybe I'll discover anything else. Anyways I'll get to you in a bit.

simenandre:
> Hi @simenandre,
> 
> 
> 
> from a quick glance at #755 I believe this could achieve the same goal, while my solution approach gives a more granular access, because the `id` could be any string, rather than only the stack itself. But I would agree there might not be a use case with two comments using the same stack.
> 
> 
> 
> I'll checkout the latest release and test it for our needs. Maybe I'll discover anything else. Anyways I'll get to you in a bit.

Thanks! 🙌

philprime:
Sorry for getting back to you so late.
We moved away from our fork and are now also using the official `pulumi/actions` step again.
Therefore I'll close this PR

PR #810, opened 2022-12-23
feat: added support to install manually install plugins

philprime:
This adds the option to manually install Pulumi plugins.

We are currently using [this Sentry plugin](https://github.com/pulumiverse/pulumi-sentry) and repeatedly noticed the following issue:

```
Logging into the Pulumi Service backend.
pulumi preview on shared
  Error: code: -2
   stdout: 
   stderr: Command failed with exit code 255: pulumi preview --parallel 2147483647 --exec-agent pulumi/actions@v3 --color auto --exec-kind auto.local --event-log /tmp/automation-logs-preview-7OA2LT/eventlog.txt --stack shared --non-interactive
  error: could not load plugin for sentry provider 'urn:pulumi:shared::***-cloud::pulumi:providers:sentry::default_0_0_8_github_/api.github.com/pulumiverse': Could not automatically download and install resource plugin 'pulumi-resource-sentry' at version v0.0.8, install the plugin using `pulumi plugin install resource sentry v0.0.8`.
  Underlying error: error downloading plugin sentry to file: failed to download plugin: sentry-0.0.8: 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-sentry-v0.0.8-linux-amd64.tar.gz
  Previewing update (shared)
  
  View Live: https://app.pulumi.com/***/***-cloud/shared/previews/1e8ae06b-b2c3-49f0-b91f-db8ab01[35](https://github.com/kula-app/kula-mono-repo/actions/runs/3764563516/jobs/6399153978#step:11:36)54b

  @ Previewing update..........
```

As the log also states, as automatically installing the plugin failed, we are changing our workflow to manually install this specific plugin. Due to action re-installing the pulumi CLI in our Docker-based GitHub Actions, we want to install the plugins right before running other commands. *EDIT: after looking further into the code, I realized only `~/.pulumi/bin` is deleted by this step, and not the full pulumi installation folder*

Therefore this adds the option to declare a list of plugins to manually install.
As nested array objects are not supported by GitHub Actions, we decided to use a YAML-string and parse it at runtime instead.

PR #812, opened 2022-12-28
feat: add plugins installation

Moon1706:
This PR allows you to install plugins from an official or your own server. As I can see in practise sometimes build-in forced Pulumi plugin installation doesn't work.


cobraz:
Hello @Moon1706! Thank you for opening this PR.

Recently, #810 was opened. From what I can understand, this is the same as what you're trying to add here. Can you confirm?

Moon1706:
@cobraz 
Exactly, but my solution is more suitable)

Moon1706:
@cobraz I'm ready

Moon1706:
@RobbieMcKinstry could you also check it? Of course, if you'll have time for that. Previous discussion was really useful and super nice for me)

cstrempfer:
#810 was closed, because the solution here seems to be cleaner. Can this PR be completed?

RobbieMcKinstry:
Hello! I'm afraid I'm no longer an OSS maintainer for Pulumi, so I can't push this PR forward. Best of luck, and thanks for your continued use of Pulumi!

iwahbe:
@Moon1706 It would be helpful if you let us know what problems you are hitting. We would like to ensure that `pulumi up` "just works", so the best outcome for us would be to fix the problem at the source. 

The next option that comes to mind is invoking `pulumi install` in the GitHub action, instead of adding a `plugins` JSON blob to the interface.

PR #813, opened 2022-12-28
refactor: improve config

cobraz:


RobbieMcKinstry (approved):
LGTM! I don't have any color commentary to add to this, I'm afraid.

mikocot:
it looks like this might have broken passing optional inputs, what was the goal and scope of the refactoring?

Issue #815, opened 2022-12-29
Seeing sporadic 403s when action tries to install plugins


jbrunton:
### What happened?

I'm using this action to automate deployments in a few projects. In recent days I've started seeing inconsistent failures when pulumi tries to install plugins (especially the `synced-folder` plugin, but I'm unsure if it's specific to that).

The error looks like this, apparently caused by a 403 when trying to download the plugin:

```
  error: could not load plugin for synced-folder provider 'urn:pulumi:staging::auth0-test::pulumi:providers:synced-folder::default_0_0_9': Could not automatically download and install resource plugin 'pulumi-resource-synced-folder' at version v0.0.9, install the plugin using `pulumi plugin install resource synced-folder v0.0.9`.
  Underlying error: error downloading plugin synced-folder to file: failed to download plugin: synced-folder-0.0.9: 403 HTTP error fetching plugin from https://get.pulumi.com/releases/plugins/pulumi-resource-synced-folder-v0.0.9-linux-amd64.tar.gz
```

However, the behaviour is inconsistent, and a re-run of the workflow usually fixes it.

### Steps to reproduce

It doesn't happen all the time so it may be difficult to provide a reproducible example – though I'm happy to try to create a simplified app if it's helpful – but it is happening pretty regularly. You can see an example here:

https://github.com/jbrunton/auth0-test/actions/runs/3802920177/jobs/6468891986

You can also see that attempt #2 was successful:

https://github.com/jbrunton/auth0-test/actions/runs/3802920177/jobs/6468979717

That is, the code works as is – it just occasional fails.

Given the nondeterminism, I'm wondering if I might be encountering rate limiting? That's just a guess, however.

### Expected Behavior

For the app to be provisioned.

### Actual Behavior

The provisioning fails and I have to re-run the workflow.

### Output of `pulumi about`

```
CLI
Version      3.50.2
Go Version   go1.19.4
Go Compiler  gc

Plugins
NAME           VERSION
aws            5.24.0
docker         3.6.1
nodejs         unknown
random         4.8.2
synced-folder  0.0.9

Host
OS       darwin
Version  13.1
Arch     arm64

This project is written in nodejs: executable='/Users/john/.nvm/versions/node/v16.19.0/bin/node' version='v16.19.0'

Current Stack: jbrunton/auth0-test/dev

TYPE                                               URN
pulumi:pulumi:Stack                                urn:pulumi:dev::auth0-test::pulumi:pulumi:Stack::auth0-test-dev
[...]

PR #818, opened 2023-01-02
docs: set GITHUB_TOKEN to avoid rate limits

jbrunton:
I realised this is _likely_ the easiest fix for #815. (I imagine also related to #810 and #812.)

I say _likely_ because it's a little bit difficult proving a fix for something which only occasionally failed, but in my testing this action has been successful in about a dozen recent runs after setting the GITHUB_TOKEN variable, whereas before I was seeing failures every 3-4 runs.

I'm fairly confident this is the correct fix, however:

1. The failures were related to downloading plugins, which are hosted on GitHub.
    * Although the error message would reference `get.pulumi.com` (e.g. `https://get.pulumi.com/releases/plugins/pulumi-resource-synced-folder-v0.0.9-linux-amd64.tar.gz`), it seems that the Pulumi retry mechanism for downloading plugins isn't robust and sometimes falls back to the wrong server (see discussion in https://github.com/pulumi/pulumi/issues/11743).
2. GitHub returns a 403 when rate limiting.
3. We see 403s frequently in CI while using this action. This makes sense: GitHub's rate limits would be shared across all users.
4. Pulumi does use the GITHUB_TOKEN when it's set ([link](https://github.com/pulumi/pulumi/blob/fc607c277c5166bf9ee78a88803d05c43849dc92/sdk/go/common/workspace/plugins.go#L251)). But by default, this action doesn't encourage this (or set it automatically).


README.md
```diff
@@ -115,6 +116,10 @@ By default, this action will try to authenticate Pulumi with the
 `PULUMI_ACCESS_TOKEN` then you will need to specify an alternative backend via
 the `cloud-url` argument.
 
+If you are using any Pulumi plugins then you should also set the `GITHUB_TOKEN`
+environment variable in order to avoid hitting GitHub rate limits. You can set
+it to the default `${{ secrets.GITHUB_TOKEN }}`.
```
cobraz, line 121: I guess users need to set a personal GitHub-token?

jbrunton: No, the GITHUB_TOKEN secret is always added by GitHub automatically: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#about-the-github_token-secret

However, they would have the option of using another token if they wish.

jbrunton: I can clarify that in the text, however. And a link to the above wouldn't hurt.

jbrunton: @cobraz: I also just realised it's possible the action could automatically set the GITHUB_TOKEN variable in the process before invoking pulumi, if users don't set it themselves. Actions should always be able to access the `github.token` context. I've not looked into this, however.

Moon1706: @jbrunton 
Will this approach work for GitHub Enterprise?

jbrunton: @Moon1706 : I don't use GitHub Enterprise so I can't verify it myself, but I don't see any reason why not: `${{ secrets.GITHUB_TOKEN }}` should always be set, even for Enterprise repos. (This is explicitly mentioned in the [docs](https://docs.github.com/en/actions/security-guides/automatic-token-authentication#how-the-permissions-are-calculated-for-a-workflow-job).)

1oglop1: Maybe also add this link about GITHUB_TOKEN and rate limits as it mentions GH Enterprise https://docs.github.com/en/rest/overview/resources-in-the-rest-api?apiVersion=2022-11-28#requests-from-github-actions

jbrunton: @1oglop1: Thanks for the suggestion. I updated the PR with that link.

jbrunton:
Btw, I realise the Pulumi docs for this action would also need updating: https://github.com/pulumi/pulumi-hugo/blob/master/themes/default/content/docs/guides/continuous-delivery/github-actions.md

I would be happy to contribute a PR for that too, but I will wait for feedback on this PR first.

README.md
cobraz: We already set the Github-token, it's used for pull request comments. We can easily reuse that.

jbrunton: @cobraz: Something like [this](https://github.com/pulumi/actions/compare/master...jbrunton:pulumi-actions:use-github-token) instead?

I'd be happy to contribute a PR, but would appreciate a design review first. I couldn't see a better way to pass the token to Pulumi but the above change works.

Moon1706:
@jbrunton 
@cobraz 
Unfortunately, adding GITHUB_TOKEN env didn't help for GitHub Enterprise server.

Moon1706:
@cobraz 
Could you add more stable functionality for Enterprise solutions? I mean [this](https://github.com/pulumi/actions/pull/812) or [this](https://github.com/pulumi/actions/pull/810) PR?

Issue #674, opened 2022-07-04
Question: getVersionObject is it possible to turn it off?


bitdevel:
### What happened?

In our environment we do not want to have outgoing traffic to the internet but in` https://raw.githubusercontent.com/pulumi/actions/master/dist/index.js` we see `function downloadCli` that calls `getVersionObject`.  Is it possible to somehow turn it off?

### Steps to reproduce

n/a

### Expected Behavior

no outbound traffic to the internet

### Actual Behavior

outbound traffic to the internet

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

cobraz:
Thanks, @bitdevel for posting this!

You're right, we should probably add some way to deactivate the functions that download stuff. How are you solving downloading the pulumi cli, just have it downloaded on the runner?

mikhailshilkov:
Yes, @bitdevel, could you please describe your target scenario for version/download resolution? That function is used to fine a version of pulumi and its download location, how would you want to override it? Some sort of environment variables?

RobbieMcKinstry:
@bitdevel just wanted to follow up on your question. 

Would you mind taking a minute to describe how you've installed the Pulumi CLI if not using the outbound network? And how you'd prefer version resolution to be performed? I can image an `offline=true` setting, but I'm curious about the edge cases others have asked about.

> could you please describe your target scenario for version/download resolution? That function is used to find a version of pulumi and its download location, how would you want to override it? Some sort of environment variables?

Thanks for your time and input! It's much appreciated as we iterate. :D

RobbieMcKinstry:
Another consideration: an `offline` mode is incompatible with... most backends. You could use the file state backend but not a cloud storage backend or the Pulumi service. Not sure how we'd handle that.

bitdevel:
Basically in our  networking setup we do not have accesss to the internet. Pip repository is mirrored in our network, we also have rules allowing to access AWS s3. This specific function tries to go directly to the internet.

cobraz:
> Basically in our networking setup we do not have accesss to the internet. Pip repository is mirrored in our network, we also have rules allowing to access AWS s3. This specific function tries to go directly to the internet.

In that case, one solution could be to open up this:
https://raw.githubusercontent.com/pulumi/docs/master/data/versions.json

RobbieMcKinstry:
[...]

PR #837, opened 2023-01-16
Skip Pulumi CLI download if runner version satisfies requested CLI version

### LINKED ISSUE #789: Optimization: Skip Download Pulumi CLI when already present
## Hello!

## Issue details

With the exception of MacOS and self-hosted runners, Pulumi comes installed on GitHub Actions runners by default. Right now, when a user does not specify a specific version of Pulumi to use, this action defaults to `latest`. Instead, we should use whatever version of Pulumi is installed on the runner, if one exists. This will remove the need to re-download Pulumi for the vast majority of users (anyone who is using Windows or Ubuntu runners).

As described in https://github.com/pulumi/actions/issues/787, this would eliminate `~100MBs` of network traffic. And, detailed in https://github.com/pulumi/actions/issues/674, it would allow the action to be used in the context of where outbound traffic is blocked.

### Affected area/feature

[The `downloadCli` function.](https://github.com/pulumi/actions/blob/master/src/libs/pulumi-cli.ts#L34)

ehafenmaier:
- Add functionality to check GitHub runner for the existence of the Pulumi CLI. 
- If the CLI is already installed on the runner check if the runner CLI version matches the requested version from the action input.
- If runner CLI version matches, skip downloading CLI.
- If CLI not installed on runner or runner CLI does not match requested version, proceed to download CLI

```mermaid
flowchart LR
    A(Action Start) -- Check Runner CLI --> B{CLI Installed\n On Runner?}
    B -- Yes --> C{Runner CLI\n Version Match?}
    C -- Yes --> D(Skip CLI Download)
    C -- No --> E(Download CLI)
    B -- No --> E
```

> NOTE: Skipping the CLI download will also skip the outbound network call in the `getVersionObject()` method, which was mentioned in #674 

This resolves #789


src/libs/pulumi-cli.ts
```diff
@@ -42,6 +61,23 @@ export async function downloadCli(range: string): Promise<void> {
...
 
+  // Check for version of Pulumi CLI installed on the runner
+  const runnerVersion = await getVersion();
+
+  if (runnerVersion) {
+    core.info(`Runner version: ${runnerVersion}`);
+    // Check if runner version matches
+    if (semver.satisfies(runnerVersion, range)) {
+      // If runner version matches, skip downloading CLI by exiting the function
+      core.info(`Runner version ${runnerVersion} matched. Skipping Pulumi CLI download`);
```
RobbieMcKinstry, line 72:
```suggestion
      core.info(`Pulumi version ${runnerVersion} is already installed on this machine. Skipping downloading.`);
```

src/libs/pulumi-cli.ts
```diff
@@ -42,6 +61,23 @@ export async function downloadCli(range: string): Promise<void> {
...
+      return;
+    } else {
+      core.info('Pulumi CLI on the runner does not match. Proceeding to download');
```
RobbieMcKinstry, line 75:
```suggestion
      core.info('Pulumi ${runnerVersion} does not satisfy the desired version ${range}. Proceeding to download');
```

src/libs/pulumi-cli.ts
```diff
@@ -42,6 +61,23 @@ export async function downloadCli(range: string): Promise<void> {
...
+    }
+  } else {
+    core.info('Pulumi CLI not installed on the runner. Proceeding to download');
```
RobbieMcKinstry, line 78:
```suggestion
    core.info('Pulumi is not detected in the PATH. Proceeding to download');
```

src/libs/pulumi-cli.ts
```diff
@@ -12,6 +12,25 @@ export async function isAvailable(): Promise<boolean> {
...
+  core.debug(`StdErr: ${res.stderr}`)
+
+  // Only check for success and if the [stdout] starts with the version
+  // prefix 'v'. If success is true and the runner version is not the 
+  // latest version then the "warning of newer version" will be in [stderr] field, 
+  // which will trigger an else condition if we also check for [stderr === '']
+  if (res.success && res.stdout.startsWith('v'))
+    return res.stdout.substring(1); // Return version without 'v' prefix
+  else
+    return undefined;
```
RobbieMcKinstry, line 31:
```suggestion
  if (res.success && res.stdout.startsWith('v')) {
    // Return version without 'v' prefix
    return res.stdout.substring(1);
  } else {
    return null;
  }
```

src/libs/pulumi-cli.ts
```diff
@@ -12,6 +12,25 @@ export async function isAvailable(): Promise<boolean> {
   return res.stderr != '' && !res.success ? false : res.success;
 }
 
+export async function getVersion(): Promise<string | undefined> {
```
RobbieMcKinstry, line 15: My internal style-guide is that undefined should be reserved for instances where a variable is not in scope, while null is intended to convey the absence of a defined value.

That's my preference, but other philosophies might be better informed than mine!

```suggestion
export async function getVersion(): Promise<string | null> {
```

src/libs/pulumi-cli.ts
```diff
@@ -12,6 +12,25 @@ export async function isAvailable(): Promise<boolean> {
...
+  core.debug('Executing command \'pulumi version\' to check version of CLI on the runner');
+  
+  const res = await exec.exec('pulumi', ['version'], true);
+
+  core.debug(`Success: ${res.success}`)
+  core.debug(`StdOut: ${res.stdout}`)
+  core.debug(`StdErr: ${res.stderr}`)
```
RobbieMcKinstry, line 22: Let's either drop these log lines, or provide a little more context about where they're running so they can be more easily differentiated from similar lines during debugging. If they were useful to you while debugging, I imagine they'd be useful to us too.

src/libs/pulumi-cli.ts
```diff
@@ -42,6 +61,23 @@ export async function downloadCli(range: string): Promise<void> {
 
   core.info(`Configured range: ${range}`);
 
+  // Check for version of Pulumi CLI installed on the runner
+  const runnerVersion = await getVersion();
+
+  if (runnerVersion) {
+    core.info(`Runner version: ${runnerVersion}`);
```
RobbieMcKinstry, line 68:
```suggestion
```

RobbieMcKinstry (changes requested):
Hey @ehafenmaier another great submission! Thank you very much for your diligent contribution!

I have a few nit-picks around the log text but otherwise it looks great! 
Lastly, would you add an entry to the changelog describing this feature? After that, we can merge :)

src/libs/pulumi-cli.ts
ehafenmaier: Even though they did help me debug the `ExecResult` and figure out what to use in the `if` statement, I'm fine with removing them. I started with your suggestion of checking `stderr === ''` but discovered that didn't work using the debug logs. I made sure to add a comment block explaining that so I think the debug logs have served their purpose and are not needed anymore.

src/libs/pulumi-cli.ts
dixler: Internally, the service team follows https://github.com/microsoft/TypeScript/wiki/Coding-guidelines

https://github.com/microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined

undefined is preferred over null

RobbieMcKinstry: Well, judging by the H1s at the top of the document, I'm inclined to disregard advice without any reasoning provided ;D

Frassle: undefined is the norm for typescript here. The optional type `string?` would expand out to `string | undefined`. The TS tooling probably understands that style better.

ehafenmaier: I fully admit that TypeScript is not my strongest language so I let the tooling guide me.  I'm coming from a C# background so I started with a function return type of `Promise<string?>`. The TS compiler complained with the following message:
```
// JSDoc types can only be used inside documentation comments.(8020)
```
During my investigation into this message I ran across this [blog post](https://paulsebastian.codes/why-isnt-string-allowed-as-a-function-return-value-in-typescript) which laid out a pretty good explanation on why `string?` is considered an invalid return type.

I'm happy to follow whatever style guide is decided upon.

Issue #804, opened 2022-12-21
Cache pulumi plugins and policy packs


RobbieMcKinstry:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

Originally opened by @josealban in [setup-pulumi](https://github.com/pulumi/setup-pulumi/issues/39):
Jose writes:

>  Pulumi plugin download can take awhile - it would be great if this action optionally cached (disabled by default) the plugins directory and up/downloaded via github artifact action

### Affected area/feature

Actions cache.

RobbieMcKinstry:
We could start by caching the CLI binary itself.

RobbieMcKinstry:
I'm starting to implement this.

RobbieMcKinstry:
[Here's my crude first pass.](https://github.com/pulumi/actions/tree/mckinstry/cache-plugins)
I'm unsure if we need to add the tools cache to the path at the beginning of execution, or perform some other action to rehydrate executables from the cache. 

Also, it would also be wise to include a test to validate cache hits vs cache misses, but I'm not quite sure how to do that.

Another approach would be to have a second executable (`post-execution.ts`) that caches plugins after execution, only if the status code was successful. This would decouple caching from the main execution loop, and I think it's more idiomatic for GitHub Actions.

JoseAlban:
Thanks for starting on this 😄 - but sad we still don't have it merged 😿

raphaelauv:
```yaml
jobs:
  pulumi:
    name: "Pulumi"

    runs-on: ubuntu-latest

    env:
      PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
      PULUMI_STACK: YYYYYYY

    steps:
      - uses: actions/checkout@v6

      - uses: actions/setup-python@v6
        with:
          python-version: 3.11

      - uses: astral-sh/setup-uv@v8.2.0
        with:
          enable-cache: true
          cache-suffix: "pulumi"
          python-version: 3.11

      - name: Cache Pulumi plugins and policy packs
        uses: actions/cache@v6
        with:
          path: |
            ~/.pulumi/plugins
            ~/.pulumi/policies
          key: ${{ runner.os }}-pulumi-${{ hashFiles('pulumi/uv.lock') }}
          restore-keys: |
            ${{ runner.os }}-pulumi-

[...]

PR #841, opened 2023-01-17
Cache Plugins on Runner between Runs

RobbieMcKinstry:
This PR caches downloaded plugins between runs as a performance optimization.

Closes https://github.com/pulumi/actions/issues/804


bennettp123:
For what it's worth, even if the plugin caching isn't working, I'd be keen for the install-only workflow on its own.

PR #851, opened 2023-01-21
Check the pulumi installed is what get-version resolved to

Frassle:
Windows wasn't using the pulumi bin that get-version resolved to, so with just the new error check would fail all runs. So this also fixes the windows install so it actually uses the pulumi downloaded rather than the default pulumi installed on GHA runners.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@
 
 _(none)_
 
---
+- bug: Fix installation on Windows.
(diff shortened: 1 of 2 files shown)


Frassle:
Note that it looks like https://github.com/pulumi/setup-pulumi needs a similar fix.

RobbieMcKinstry (approved):

PR #852, opened 2023-01-21
testing empty config

Frassle:


Frassle:
Repro'd in pu/pu with just the automation api: https://github.com/pulumi/pulumi/pull/11943
So clearly not an action bug.

PR #859, opened 2023-02-04
strip ansi control chars from PR comments

bennettp123:
When `pulumi/actions` is used with `{comment-on-pr: true, color: "always"}`, it creates a PR comment with extra ANSI control characters.

Let's strip them before adding them to the PR comment.

Before:

image

After:

image

Uses `ansi-to-html` to convert the output into HTML. The result is then wrapped in `<pre></pre>` tags before adding it to the PR comment.

### DIFF
--- a/src/libs/pr.ts
+++ b/src/libs/pr.ts
@@ -1,8 +1,22 @@
 import * as core from '@actions/core';
 import { context, getOctokit } from '@actions/github';
-import * as dedent from 'dedent';
+import AnsiToHtml from 'ansi-to-html';
+import dedent from 'dedent';
 import invariant from 'ts-invariant';
 import { Config } from '../config';
+
+function ansiToHtml(
+  message: string,
+  { maxLength }: { maxLength?: number } = { maxLength: undefined },
+) {
+  const convert = new AnsiToHtml();
+  let html = convert.toHtml(message);
+  while (maxLength !== undefined && html.length > maxLength) {
+    message = message.substring(0, message.length - 1);
+    html = convert.toHtml(message);
+  }
+  return html;
+}
 
 export async function handlePullRequestMessage(
   config: Config,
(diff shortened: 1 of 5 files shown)


src/libs/pr.ts
```diff
@@ -1,9 +1,23 @@
...
+function ansiToHtml(
+  message: string,
+  { maxLength }: { maxLength?: number } = { maxLength: undefined },
+) {
+  const convert = new AnsiToHtml();
+  let html = convert.toHtml(message);
+  while (maxLength !== undefined && html.length > maxLength) {
+    message = message.substring(0, message.length - 1);
+    html = convert.toHtml(message);
+  }
```
RobbieMcKinstry, line 17: Could you explain in a comment what this code is doing? It looks like it's truncating the message body according to the max length of the message.

However, it looks like it's only cutting out a single character at a time and attempting to convert to HTML again. I imagine that would be very slow if html.length >>>> maxLength.

src/libs/pr.ts
```diff
@@ -20,7 +34,7 @@ export async function handlePullRequestMessage(
 
   const summary = '<summary>Pulumi report</summary>';
 
-  const rawBody = output.substring(0, 64_000);
+  const rawBody = ansiToHtml(output, { maxLength: 64_000 });
```
RobbieMcKinstry, line 37: Since this PR touches this code, would you mind pulling this value into a constant so it's not a ["magic number"](https://codeburst.io/software-anti-patterns-magic-numbers-7bc484f40544)? I'd appreciate it :D

RobbieMcKinstry (changes requested):
On the whole this change LGTM. I do wonder if it makes sense to add `pre` tags in all cases. My only concern is if this will affect users who have `color` disabled.

Does GitHub Flavored Markdown always convert code blocks (\`\`\`) to `<pre>` tags when it renders in the browser? i.e. is change equivalent when `color: never`?

PR #860, opened 2023-02-04
Skip Pulumi CLI download if runner version satisfies requested CLI version

### LINKED ISSUE #789: Optimization: Skip Download Pulumi CLI when already present

ehafenmaier:
Accidentally closed PR #837 when syncing fork. 

## Original PR
- Add functionality to check GitHub runner for the existence of the Pulumi CLI. 
- If the CLI is already installed on the runner check if the runner CLI version matches the requested version from the action input.
- If runner CLI version matches, skip downloading CLI.
- If CLI not installed on runner or runner CLI does not match requested version, proceed to download CLI

> NOTE: Skipping the CLI download will also skip the outbound network call in the `getVersionObject()` method, which was mentioned in #674 

This resolves #789


src/libs/pulumi-cli.ts
```diff
@@ -43,6 +57,22 @@ export async function downloadCli(range: string): Promise<void> {
...
   core.info(`Configured range: ${range}`);
 
+  // Check for version of Pulumi CLI installed on the runner
+  const runnerVersion = await getVersion();
+
+  if (runnerVersion) {
+    // Check if runner version matches
+    if (semver.satisfies(runnerVersion, range)) {
+      // If runner version matches, skip downloading CLI by exiting the function
+      core.info(`Pulumi version ${runnerVersion} is already installed on this machine. Skipping downloading.`);
```
Frassle, line 67:
```suggestion
      core.info(`Pulumi version ${runnerVersion} is already installed on this machine. Skipping download`);
```

src/libs/pulumi-cli.ts
```diff
@@ -43,6 +57,22 @@ export async function downloadCli(range: string): Promise<void> {
...
+    // Check if runner version matches
+    if (semver.satisfies(runnerVersion, range)) {
+      // If runner version matches, skip downloading CLI by exiting the function
+      core.info(`Pulumi version ${runnerVersion} is already installed on this machine. Skipping download`);
+      return;
+    } else {
+      core.info(`Pulumi ${runnerVersion} does not satisfy the desired version ${range}. Proceeding to download`);
+    }
+  } else {
+    core.info('Pulumi is not detected in the PATH. Proceeding to download');
```
cobraz, line 73: Maybe we should use debug here since we're already sending so many info logs?

```suggestion
    core.debug('Pulumi is not detected in the PATH. Proceeding to download');
```

cobraz (approved):

src/libs/pulumi-cli.ts
ehafenmaier: Seems like if this last log message is switched to debug the other two should follow suit since they're all part of the same set of conditions (the outer if and the inner if). Only one of the three will be written out so I guess the question is _are these messages helpful in general or should they only be used for troubleshooting?_

cobraz (approved):

PR #865, opened 2023-02-09
Add a diff summary above the Pulumi output in PR comments

JDTX0:
This PR adds a brief summary to PR comment just before the Pulumi output. It includes a URL to view the preview, and a brief summary of all changes.

I think it's useful to see an overall view without having to expand the full output and scroll to the bottom.

With these changes, this is what the new output should look like:
Screenshot 2023-06-13 at 2 41 14 PM

My method of finding the diffs and URL involves grabbing the first & last 500 chars of the Pulumi output, then using simple regex to search for changes/additions/deletions among the last 500 chars. This approach should take minimal resources.

If there's a better way to accomplish this or if I've missed something in my changes, please let me know.

Thanks!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 --
 
+## 4.5.0 (2023-06-13)
+
+- feat: Show a brief summary and direct URL to Pulumi UI in PR comments
+
 ## 4.4.0 (2023-06-05)
 
 - feat: Download CLI if preinstalled version has a known issue.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,7 +86,7 @@
 
 - enhancement: Add `pulumi-version` option to allow pinning the version of the
   CLI [PR](https://github.com/pulumi/actions/pull/661) fixes
-  [#437 ](https://github.com/pulumi/actions/issues/437)
+  [#437](https://github.com/pulumi/actions/issues/437)
 
 ## 3.18.1 (2022-07-07)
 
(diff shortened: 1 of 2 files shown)


justinvp:
Sorry for the long delay @JDTX0! We missed this in our daily triage of incoming issues/PRs.

Definitely agree that it would be useful to have a summary here! Although, I think we'll want to iterate on the design a bit here, likely aligning it more closely with the output of the [GitHub App](https://www.pulumi.com/docs/using-pulumi/continuous-delivery/github-actions/#pulumi-github-app). I've opened an issue to track the overall suggestions here: https://github.com/pulumi/actions/issues/969

Implementation wise, rather than using regex to extract the data from the raw output, we could use the automation API operation's result, which includes a [map of operation types (e.g. delete/update/create/etc.) along with corresponding counts](https://github.com/pulumi/pulumi/blob/1befe083d8f0e911ef6f166fa5b5982feab50030/sdk/nodejs/automation/stack.ts#L799-L804). We'd have to modify the code here to return that summary information:

https://github.com/pulumi/actions/blob/f9775fbf93748847db8f2cf9f9ed889f37de416c/src/main.ts#L87-L101

For getting the link, we do have [an API in the Go Automation API](https://github.com/pulumi/pulumi/blob/34f524b34d1a9a99872e72519c66748939688c33/sdk/go/auto/stack.go#L897-L900) for retrieving this, but haven't added it to the other languages yet. This is tracked by https://github.com/pulumi/pulumi/issues/5533. We should add support to the Node.js automation API and make use of that to grab the permalink URL.

JDTX0:
@justinvp Good points, didn't realize there was an automation API to retrieve those.

I'll close this out and let the Pulumi team handle implementation. Or if I get enough time one of these days then I'll open a PR that utilizes the automation API.

Thank you for the response & opening the feature request. much appreciated!

Issue #868, opened 2023-02-16
Support for 'stack output' command


mikocot:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

- Add `output` command support to actions

There are multiple scenarios where it's useful to fetch outputs from a stack without having to first call `preview` or `up`.. Reading outputs is fast and has no side effects. while lets you access stack configuration that may not be available anywhere else.

For now we are running `pulumi stack output` as bash code in our pipeline, but this is ugly and hard to maintain. It's also especially annoying as the whole logic to pass outputs and secrets into github workflow is already there and triggered as part of calling `up`/`preview`/`refresh`/.... All it takes is to allow that same logic by itself.

### Affected area/feature

I have looked into solving it myself and it looks like a very subtle change. Please see:
[support for stack output](https://github.com/mikocot/pulumi-actions/commit/a6b265e8c6f22edd930417e5ca5452fb2b45d700)

Yet, for some reason the outputs don't seem to be rendered and I don't really have the test environment and sufficient understanding of underlying system to investigate it. I guess some preconditions might need to be met before `stack.outputs()` is called or some configuration needs to be adapted.

mikocot:
TBH I dont see much wrong about the action becoming a wrapper for the CLI =D Since it only runs the command behind the scenes it feels there is not much duplication and maintenance needed. Anyway, this particular command surely has a value for CI/CD.

As for the implementation, maybe it's just me testing it incorrectly. I'll make a PR and you can check it yourself.

juchom:
I'm playing a bit with Pulumi and Kubernetes with github actions and I just this isse

It would be great to have this, because in my case, if nothing has changed in my pulumi folder I skip the pulumi job.

Being able to access pulumi secrets / outputs in other jobs without a preview or up would be great.

mikocot:
we've just found some more scenarios where we need to construct a pretty ugly and error pron bash script to get the outputs, while this feature seems to be quite close to finishing with my PR, any chances we'll see it merged soon?

RobbieMcKinstry:
Hi Mikocot,
Sorry, I haven't had an opportunity to revisit some of the PRs on this repo just yet. I'm going to try to make some time next week. Apologies for the slow reply!

mikocot:
Looking at the likes it's not only our project that would benefit from this one :) The PR is still up for merge

[...]

PR #869, opened 2023-02-16
Support for 'stack output' command

mikocot:
This PR introduced support for `pulumi stack output` command in addition to the existing ones. 

The reasons for this change are explained in the following issue:
[https://github.com/pulumi/actions/issues/868](https://github.com/pulumi/actions/issues/868)

I've chosen `output` as a name for the command as it reflects the method called by CLI, although internally in the sdk it calls `stack.outputs()`.

As mentioned previously I have issues testing this solution, so I invite you to test it yourself and collaborate on the final solution. I have found no relevant tests nor I can see a good way to unit test it, but I've updated the test workflow.


src/main.ts
```diff
@@ -88,6 +89,7 @@ const main = async () => {
 
   for (const [outKey, outExport] of Object.entries(outputs)) {
     core.setOutput(outKey, outExport.value);
+    onOutput(outKey + ' found');
```
RobbieMcKinstry, line 92: Was this line added as part of the debugging process? I'm not totally able to make sense of it's intent. Forgive me if it should be obvious. 

I'm going to checkout this PR locally and see if I can add a test. I'll take a closer look locally.

RobbieMcKinstry (changes requested):
Just one question; otherwise, it looks good. I think `output` captures the command's functionality. I'd like a test and additional documentation in the README, but otherwise it looks good.

I'm going to check out this code locally and see if I can add a test.

RobbieMcKinstry:
Oh, also, it looks like this PR requires a rebase before we can merge it.

src/main.ts
mikocot: yeah, probably, it's been a while :)

mikocot:
The PR has been reopened as #1002 

https://github.com/pulumi/actions/pull/1002

@RobbieMcKinstry could we please have it behind us finally?

RobbieMcKinstry:
Hello! I'm afraid I'm no longer a maintainer of this repository. Apologies, I wish I could help you.

PR #903, opened 2023-04-07
Add dist files to gitignore and forcibly commit them instead

cobraz:


RobbieMcKinstry (approved):
Nice! Saves us a lot of headaches fixing up PRs that accidentally committed these files.

PR #912, opened 2023-04-11
feat: support for config on multiple types on keys and values

### LINKED ISSUE #911: config-map failing with "Expected string, but was null" in v4.2
### What happened?

Pipeline failures after release of v4.2. 

Pipeline code:
```
 uses: pulumi/actions@v4
[...]
 config-map: | # provide required configuration
   { prNumber: { value: ${{ github.event.pull_request.number }} } }
```

Error message
`Validation failed: { "prNumber": { "value": "Expected string, but was null" } }. Object should match { [_: string]: { value: string; secret?: boolean; } }`

The same happens if `${{ github.event.pull_request.number }}` is provided, but it's a number, the error is then:

` "Expected string, but was number"`

### Expected Behavior

No breaking changes in minor release.
Changes in interface communicated in release notes and PR descriptions

### Steps to reproduce

See code above

### Output of `pulumi about`

not essentiral here

### Additional context

The issue can be fixed by wrapping params in double quotes or ensuring it's a string.

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

simenandre:
Before #813 we naively accepted all kinds of objects into `config-map`, which has [led to users having other types](https://github.com/pulumi/actions/issues/911) than what we [have defined in our restrictive runtypes](https://github.com/pulumi/actions/blob/710a2810bf92e422f1aafd72323e019cfd7672b0/src/config.ts#L30-L36).

I think what is described in #911 should be supported, and as @mikocot [describes in a comment](https://github.com/pulumi/actions/issues/911#issuecomment-1503011251) there are probably a lot of users that depend on that. Since we're currently reflecting the types that are supported by `stack.setAllConfig` from the Automation API, it need to be [extended to support more types](https://github.com/pulumi/pulumi/issues/12641). IMHO, it should be fine to merge this pull request as a temporary fix (with the [`as any` hotfix](https://github.com/pulumi/actions/blob/d606c45e60d896e399cf6bd3ed87f7554b3bf8db/src/main.ts#L76-L77) along with a TODO and issue)

fixes #911

blocked by pulumi/pulumi#12641

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## HEAD (Unreleased)
 
-**(none)**
+- feat: support for config on multiple types on keys and values
 
 --
 


simenandre:
Having thought about it for a minute, should we parse to string here?

RobbieMcKinstry:
> Having thought about it for a minute, should we parse to string here?
Can you clarify which line you're referring to?

RobbieMcKinstry:
Oh, like convert to string instead of casting to `any`?

RobbieMcKinstry:
Also, the build is red because the CHANGELOG is missing an entry.

RobbieMcKinstry:
Just thinking out loud here:

If we pass any, we're passing a strongly typed map and telling the compiler to infer the type is a string-to-string map. When a  key is accessed, if the AutomationAPI assumes its a string, then we'll have a type error.

If we convert the keys from e.g. Boolean to String, then the Automation API will be correctly typed. 
----------
I scanned the AutoAPI code to see how it handles keys. It looks like it coerses them to a string (even though they're supposed to be strings, but like it our case, users may lie to the compiler). https://github.com/pulumi/pulumi/blob/master/sdk/nodejs/automation/server.ts#L79 
 
So it looks like the `any` approach is safe, but that relies on the internals of the AutoAPI to do the right thing. I think coercing the values to a string is safer.

RobbieMcKinstry:
Ah, looks like the `getConfigMap` function I referenced points to a different `getConfigMap` (coming from protobuf)?

simenandre:
So, here's another alternative. #913 removes validation, getting us back to how we did it before #813.
Then we can put this PR on hold until https://github.com/pulumi/pulumi/issues/12641 is fixed.

PR #925, opened 2023-04-27
feat: Support `exclude-protected` on `destroy`

### LINKED ISSUE #628: Support the `--exclude-protected` on destroy
## Hello!

## Issue details

For similar reasons outlined in the [issue](https://github.com/pulumi/pulumi/issues/6539) that resulted in the `--exclude-protected` being added to the CLI, it would be good to be able to provide this flag in the GitHub Action.

justinvp:
Support passing the `--exclude-protected` flag to `pulumi destroy`. Requires the latest version of `@pulumi/pulumi` where the option is now available in Automation API (see https://github.com/pulumi/pulumi/pull/12734).

Fixes #628

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ The action can be configured with the following arguments:
   control as part of the action if you wish to perform any further tasks with
   that stack.
 
-- `exclude-protected` - (optional) Do not destroy protected resources. Only
+- `exclude-protected` - (optional) Skip destroying protected resources. Only
   valid when `command` is `destroy`.
 
 By default, this action will try to authenticate Pulumi with
--- a/README.md
+++ b/README.md
@@ -121,6 +121,9 @@ The action can be configured with the following arguments:
   control as part of the action if you wish to perform any further tasks with
   that stack.
 
+- `exclude-protected` - (optional) Do not destroy protected resources. Only
+  valid when `command` is `destroy`.
+
 By default, this action will try to authenticate Pulumi with
 [Pulumi Cloud](https://app.pulumi.com/). If you have not specified a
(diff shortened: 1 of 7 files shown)


simenandre (approved):
This looks good!

We don't need to update the CLI for this, just the SDK/NPM?

justinvp:
> We don't need to update the CLI for this, just the SDK/NPM?

The flag has been available in the CLI for a couple years now (https://github.com/pulumi/pulumi/pull/8359)

README.md
```diff
@@ -121,6 +121,9 @@ The action can be configured with the following arguments:
   control as part of the action if you wish to perform any further tasks with
   that stack.
 
+- `exclude-protected` - (optional) Do not destroy protected resources. Only
```
RobbieMcKinstry, line 124:
```suggestion
- `exclude-protected` - (optional) Skip destroying protected resources. Only
```

This change in language disambiguates the distinction between Pulumi and Terraform, where `protect` will always prevent the destruction of the resource, and `destroy` will fail under any circumstance.

action.yml
```diff
@@ -89,6 +89,10 @@ inputs:
     description: 'Colorize output. Choices are: always, never, raw, auto'
     required: false
     default: 'auto'
+  exclude-protected:
+    description: 'Do not destroy protected resources. Only valid when command is destroy.'
```
RobbieMcKinstry, line 93:
```suggestion
    description: 'Skip destroying protected resources. Only valid when command is destroy.'
```

RobbieMcKinstry (approved):
Smol language nit; otherwise LGTM

PR #956, opened 2023-06-02
feat: Download CLI if preinstalled version is known bad

### LINKED ISSUE #955: Prefer downloading the CLI if the version preinstalled on the runner is known bad
## Hello!

## Issue details

Sometimes a release of the CLI contains a bug that is bad enough that we'd prefer users avoid using that version, to avoid the issue. For example, a bug that could cause stack corruption. We'd like to have a holistic way to "taint" a bad release, but in the meantime, one tactical change could be made to the action to prefer downloading a more recent version of the CLI if the version that is preinstalled has known issues.

justinvp:
Sometimes a release of the CLI contains a bug that is bad enough that we'd prefer users avoid using that version, to avoid the issue. For example, a bug that could cause stack corruption. We'd like to have a holistic way to "taint" a bad release, but in the meantime, one tactical change is to have the action prefer downloading a more recent version of the CLI if the version that is preinstalled has known issues. For now, the list is hardcoded, but in the future it could be stored in a centralized location that the action could check.

Fixes #955

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -48,6 +48,35 @@ jobs:
     outputs:
       changed: ${{ steps.changes.outputs.dist }}
 
+  test-install-only-without-removal-of-pre-installed-pulumi:
+    needs: install-and-build
+    if:
+      ${{ needs.install-and-build.outputs.changed == 'true' || github.event_name
+      == 'workflow_dispatch' }}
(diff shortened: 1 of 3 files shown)


justinvp:
We should be able to see the results with the existing tests that run in a workflow.

dixler (approved):

simenandre (approved):

PR #987, opened 2023-07-14
chore: change to renovate

simenandre:
I changed from Dependabot to Renovate on most projects of mine a few months ago to see if that helped my workflow. I saw too many updates from Dependabot and felt stuck and overloaded. This repository is, unfortunately, no exception. 

@SimenB suggested that I'd try the same setup he has for most of Jest stuff, which essentially is exactly what I'm suggesting with this pull request. The workflow consists of:

* Maintaining lock files automatically every Monday. In our `package.json`, we typically define a package range. Exactly which package we use is updated regularly in our lock file. 
* A pull request is opened if there's an update outside of our range. Typically, we need to interact more with something outside our defined range.

I have experienced that this combination helps us pay more attention and see breaking changes.

**Note:** Renovate will open an "Dependency Dashboard" issue. This is meant to communicate and control the rate of pull requests and keep consumers of this package up to date with that, typically more than just what we have in our package.json. You can see an example here: https://github.com/simenandre/pulumi-config/issues/350 (I found a project with a lot of open PRs to showcase how that'll look like).

It supports updating all dependency files, including GitHub Actions.


simenandre:
With this latest dependabot change, I no longer this we should change to renovate.

re #1055

PR #994, opened 2023-07-18
feat: Add Update Plan Functionality

### LINKED ISSUE #991: Support Update Plans
## Hello!

## Issue details
As part of our pipeline, we do the following:

- Run unit tests against our codebase
- Do a Pulumi Preview, which then someone reviews, approves, and then the author merges.

On merge (`push`) we then run the pipeline without tests and `up` instead of `preview.` 

This presents a problem: What the reviewer signed off on is not guaranteed to be what actually gets deployed. If the `preview` showing as a PR comment is a few hours old, and someone has found a way to "clickops" something, Pulumi has the potential to undo said change (which tbh is okay.) In the worst case, it does something completely different altogether than what has been planned.

Pulumi appears to have solved this with the public preview of "[Update Plans](https://www.pulumi.com/docs/concepts/update-plans/)" but this functionality is not exposed via the Github Action.

I'd like to add this functionality to the GitHub action via adding `plan` to `config.options`. 

### Affected area/feature

Affected area would be the Github Actions repository. Would need changes to `config.ts`, `main.ts`, some tests, `README.md`, and the `examples` directory to give users a good example.

dmizelle:
Related to / resolves #991

This commit does the following:
- Updates `src/config.ts` to expose the `plan` option
- Updates the `README.md` to reflect that this option is now available.
- Adds a new test that will do a `preview`, save the plan, and then attempt to `up` using the plan.
- Update `action.yml` to reflect this new option
- `.github/workflows/workflow.yml`:
  - Adds new steps to each job for generating an Update Plan and ensuring that the file has content when testing `preview`

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -56,3 +56,51 @@ jobs:
           work-dir: .github/test-stacks/golang
           config-map: "{name: {value: test, secret: false}}"
       - run: echo 'The random string is `${{ steps.pulumi.outputs.name }}`'
+  test-update-plan: # make sure the action works on a clean machine without building
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-go@v4
+        with:
+          go-version: '^1.13.1'
+      - run: |
+          pulumi login --local
+          pulumi stack init dev
+        working-directory: .github/test-stacks/golang
+        env:
+          PULUMI_CONFIG_PASSPHRASE: not-a-secret
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+          cache: yarn
+      - run: yarn install
+      - run: yarn build
+      - name: Generate Update Plan via Pulumi Preview
+        uses: ./
+        if: always()
+        id: pulumi-preview
+        env:
+          PULUMI_CONFIG_PASSPHRASE: not-a-secret
+        with:
+          command: preview
+          cloud-url: file://~
+          stack-name: dev
+          upsert: true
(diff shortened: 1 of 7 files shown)


dmizelle-humane:
Sorry about the force-pushes, etc. Mixed up my personal + work accounts. Should be fixed now.

dmizelle:
Sorry about the force-push spam. Mixed up my personal and work accounts. Should be fixed now.

simenandre (approved):
This is just awesome, @dmizelle!

Note to self: We should also add an example where the plan is uploaded as an artifact

dmizelle:
> This is just awesome, @dmizelle!
> 
> Note to self: We should also add an example where the plan is uploaded as an artifact

> This is just awesome, @dmizelle!
> 
> Note to self: We should also add an example where the plan is uploaded as an artifact

Thank you! (and thanks for the review too :) )

I was thinking the same thing. I was sitting here trying to think of what sort of example to write. 

Our use case of uploading the artifact in a Pull Request and retrieving it during a merge to main workflow is a little complicated as the official GitHub Action for uploading an artifact doesn't support cross-workflow retrieval. Other users recommend using this one to retrieve the artifact itself:

https://github.com/dawidd6/action-download-artifact

I think I'm going to play around with this after we merge here and then submit a later pull request with an example. How does that sound?

simenandre:
> I think I'm going to play around with this after we merge here and then submit a later pull request with an example. How does that sound?

That sounds awesome!

justinvp (approved):
LGTM. Thanks, @dmizelle!

PR #1008, opened 2023-08-08
add more options to up invocation

jaxxstorm:


simenandre (commented):
Thanks Lee!

The inputs need to be added to https://github.com/pulumi/actions/blob/daa3f4a7f280f4390c9ed0f690a8607106ca8bcc/action.yml#L96-L98 as well. Would you mind adding them here as well?

We might want to test these options too, in our workflow. Debug for instance

Issue #752, opened 2022-10-11
Node.js 12 runtime for GH actions is deprecated


RobbieMcKinstry:
# Background

Node 12 was sunset on 30 Apr 2022. In non-Enterprise environments, GitHub Actions supports JavaScript-based actions running on either Node 12 or Node 16.

In GitHub Enterprise, support for Node 16 was added in GitHub Enterprise v3.4.0, released on 15 March 2022. 
We intend this action to work for all GitHub users, both SaaS users, and GitHub Enterprise users. This action will continue to use Node 12 until all maintained versions of GitHub Enterprise have Node 16 support. We don't want to break users running a maintained, supported versions of GHE that simply can't execute our action.

As a result, this Action will switch to Node 16 once GHE v3.3 is sunset. That date is 07 December 2022. At that time, all active versions of GHE will be able to execute this action. We've updated the [README](https://github.com/pulumi/actions#node-12-support-ending-2022-12-07) to reflect this support policy. 

Until 07 December, you may see the following warning:

```
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: pulumi/actions
```

![image](https://user-images.githubusercontent.com/1466314/194906101-0d61753d-5451-43b0-9664-68694deacd8f.png)

## Additional Context

For more information, please see [GitHub's deprecation notice](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) for Node 12 actions.

* [GHE Sunsetting Timeline](https://docs.github.com/en/enterprise-server@3.4/admin/all-releases#currently-supported-releases)
* [Original issue report](https://github.com/pulumi/actions/pull/651#issuecomment-1177479858)
* [Original Revert from Node 16 back to Node 12.](https://github.com/pulumi/actions/pull/678)
* https://github.com/pulumi/actions/issues/750
* https://github.com/pulumi/actions/issues/707

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

RobbieMcKinstry:
Update: it appears GHE has rescheduled the deprecation until 18 Jan. 2023. I've updated the issue text to reflect this change.

PR #1018, opened 2023-09-14
chore: node 20 as default runtime

dirien:
Hi, 

Node 16 reached the [end of life on Sep, 11 2023](https://nodejs.org/en/blog/announcements/nodejs16-eol). 

Node 20 is supported on GitHub Runners since https://github.com/actions/runner/releases/tag/v2.308.0.

This PR updates the runtime

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ---
 
+## 4.6.0 (2023-10-12)
+
+- chore: node 20 as default runtime
+
 ## 4.5.0 (2023-09-12)
 
 - feat: comment on GitHub step summary.
--- a/action.yml
+++ b/action.yml
@@ -111,5 +111,5 @@ runs:
   #
   # Trying to upgrade in July 2022 broke our users that depend on
   # GitHub Enterprise versions < 3.4
-  using: 'node16'
+  using: 'node20'
   main: 'dist/index.js'


Frassle:
Is this going to cause similar issues in GHE as the 12 to 16 bump caused? https://github.com/pulumi/actions/issues/752

simenandre:
> Is this going to cause similar issues in GHE as the 12 to 16 bump caused? https://github.com/pulumi/actions/issues/752

It might be, but I think we should think about our approach here. As far as I know, we have two alternatives.


a) Should we bump the action, potentially add a note that for runners without support for Node 16 must stay on v4.

b) Make sure most GitHub Enterprise users can use any current version of our action.

I vote for a). In general, I favor releasing often, even though it produces less compatibility. At least in cases where it's out of our hands (such as upstream stuff or GHA not supporting things). I still think we should support earlier major versions to some point, at least security issues.

What do you think?

dirien:
Bump to v5 as this will indicate the potential for breaking changes and make this clear in the release notes would be my suggestion.

justinvp:
v5 bump makes sense to me

simenandre (approved):
Let's go!

jplewa:
When will this update be released?

MatthewBettonMaif:
Hello,

Since [last published release version 4.6.0 (las hour ago)](https://github.com/pulumi/actions/releases/tag/v4.6.0) (and not a version 5 as suggested here for a breakingchange) we have now this error message :

```bash
System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter ''using: node20' is not supported, use 'docker', 'node12' or 'node16' instead.')
   at GitHub.Runner.Worker.ActionManifestManager.ConvertRuns(IExecutionContext executionContext, TemplateContext templateContext, TemplateToken inputsToken, String fileRelativePath, MappingToken outputs)
   at GitHub.Runner.Worker.ActionManifestManager.Load(IExecutionContext executionContext, String manifestFile)
 ```
 
 ```bash
 Fail to load pulumi/actions/v4/action.yml
 ```

Frassle:
Sorry about that, a long delay between that change being merged and released and no left over indicator in the changelog to remind us to bump the major version. We'll see if we can yank the 4.6 release and re-release it as 5.0.

tgummerer:
I just deleted the 4.6.0 release and the corresponding tag, so this should work again.  Thanks for reporting this @MatthewBettonMaif

MatthewBettonMaif:
Thank you for your prompt responses and action :)

I can confirm you that it is now ok

PR #1020, opened 2023-09-19
fix(examples): Updated 'setup-nodejs' actions to 'setup-node'

Jamess-Lucass:
This just updates the examples to use the correct GitHub action for setting up node. The repository is https://github.com/actions/setup-node and thus with `actions/setup-nodejs` you will receive the error `Unable to clone https://github.com/actions/setup-nodejs refs/heads/v3: repository not found`


simenandre (approved):
Thank you!

Issue #1039, opened 2023-11-21
Allow getting dev channel releases from the action


tgummerer:
Allow using dev channel releases introduced in https://github.com/pulumi/pulumi/issues/14622 by the pulumi action.  This work depends on https://github.com/pulumi/pulumi/issues/14622 and https://github.com/pulumi/pulumi/issues/14623.

/xref https://github.com/pulumi/home/issues/3003

PR #1051, opened 2023-12-20
allow getting latest dev version

tgummerer:
Add a new "dev" option for the version field, that allows users to install the latest dev release automatically.

Fixes https://github.com/pulumi/actions/issues/1039


julienp (approved):
LGTM

Issue #1014, opened 2023-09-01
error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions


hisuwh:
### What happened?

Preview pulumi build in GitHub actions on pull requests have started failing apparently for no reason.  I have tried refreshing the PULUMI_ACCESS_TOKEN

```
Error: code: -2
 stdout: 
 stderr: Command failed with exit code 255: pulumi stack select --stack org-name/stack-name --non-interactive
error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions
 err?: Error: Command failed with exit code 255: pulumi stack select --stack org-name/stack-name --non-interactive
error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions
```

### Expected Behavior

Preview build runs

### Steps to reproduce

```yaml
    - uses: pulumi/actions@v3
      with:
        command: preview
        stack-name: org-name/stack-name
        work-dir: ./infrastructure/packages/project
      env:
        PULUMI_ACCESS_TOKEN: ${{ secrets.PULUMI_ACCESS_TOKEN }}
        ARM_CLIENT_ID: ${{ secrets.AZURE_CLIENT_ID }}
        ARM_CLIENT_SECRET: ${{ secrets.AZURE_CLIENT_SECRET }}
        ARM_TENANT_ID: ${{ secrets.AZURE_TENANT_ID }}
        ARM_SUBSCRIPTION_ID: ${{ secrets.AZURE_SUBSCRIPTION_ID }}

```

### Output of `pulumi about`

```
$ pulumi about
CLI
Version      3.79.0
Go Version   go1.21.0
Go Compiler  gc

Plugins
NAME    VERSION
nodejs  unknown

Host
OS       Microsoft Windows 11 Pro
Version  10.0.22621 Build 22621
Arch     x86_64

This project is written in nodejs: executable='C:\Program Files\nodejs\node.exe' version='v18.12.1'
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Frassle:
Looks like the same as https://github.com/pulumi/actions/issues/1010.

hisuwh:
No because I'm using the Pulumi cloud backend not an Azure blob

Frassle:
[...]

PR #1062, opened 2024-01-08
return error when login fails

tgummerer:
Currently if the login fails as part of the GitHub action, we just silently ignore that.  This can be very confusing when steps further down the line fail, e.g. with "error: PULUMI_ACCESS_TOKEN must be set for login during non-interactive CLI sessions", even though the user has set the access token.

Instead of just continuing our merry way when the login fails, fail the action and let the user know.

This might be the cause of https://github.com/pulumi/actions/issues/1014.  The symptoms match, though it's unclear if there is anything else going on in that issue.

/cc @hisuwh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## HEAD (Unreleased)
 
+- fix: error out when login fails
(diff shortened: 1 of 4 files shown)


Zaid-Ajaj (approved):
LGTM

Issue #870, opened 2023-02-27
Refresh preview option


michal-bajer1:
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

I would like to build a Github workflow, which would regularly run preview of stack refresh and create an issue in the repository, if some changes between stack and our cloud provider are found.

### Affected area/feature

The options of the refresh or preview command, so that the equivalent of `pulumi preview --refresh` can be executed.

michal-bajer1:
Hi, there is a subtle difference there. 
I tried this option, but what happens is that it runs refresh, stores the results into the stack and then it runs preview. But I don't really want the state of the stack to change unpredictably. What `pulumi preview --refresh` does is effectively preview of code vs cloud provider, but it doesn't change the state of the stack.

RobbieMcKinstry:
Hi Peter (@peterstockwell-sagov),
Sorry for the slow response; we'd be happy to accept a PR for the `raw` command. :-)

> if multiple PRs are opened or updated at around the same time, one of the PR checks will fail as only one update can be performed on a stack at a time

When I read this, my mind lept to GitHub Action's [`concurrency` blocks](https://docs.github.com/en/actions/using-jobs/using-concurrency). (Forgive me if you've already considered this.) At the granularity of `jobs`, you can block jobs from running at the same time if they're both PRs to [the same `base_ref`](https://docs.github.com/en/actions/learn-github-actions/contexts#github-context), for instance. 

----

I think @michal-bajer1 identified some very confusing behavior. It's confusing (to me) that…

```yaml
- uses: pulumi/actions@v4
  command: preview
  refresh: true
```

…runs… 
```bash
$ pulumi refresh; pulumi preview
```

instead of...

```bash
$ pulumi preview --refresh
```

If I had wanted the chaining behavior, I would have thought we'd chain the operations in the YAML instead:

```yaml
- uses: pulumi/actions@v4
  command: refresh
- uses: pulumi/actions@v4
  command: preview
```

This is made even more confusing by https://github.com/pulumi/pulumi/issues/10967

Unfortunately, I don't think we have a way to repair this behavior without redefining the behavior of `refresh: true`.
That leads me to believe @peterstockwell-sagov 's suggestion of adding a `raw` command would be the way forward here.

michal-bajer1:
@RobbieMcKinstry It definitely confused me at first :) 

[...]

PR #1116, opened 2024-03-08
Allow suppressing progress dots and outputs

mikocot:
Adds support for --suppress-progress and --suppress-outputs parameters of stack up, preview, destroy and refresh.

Requires changes releasesd in the automation api (pulumi/pulumi 3.109 - https://github.com/pulumi/pulumi/pull/15596)

Original issue described here:
fixes #1108

PR #1118, opened 2024-03-10
Change `refresh` behavior to use `--refresh` flag

fitz-vivodyne:
As described in https://github.com/pulumi/actions/issues/870, the current `refresh` config option runs `pulumi refresh` prior to the requested command, updating the state file. There is not currently a way to run `pulumi preview --refresh` or `pulumi up --refresh`.

~This PR adds a `soft-refresh` flag which enables the `--refresh` flag.~
~Not sure on the right approach for naming, but I figured this helps get across the difference between `refresh` updating~ ~the state and `--refresh` not updating the state.~

After further discussion, this PR represents a breaking change. 
As of this PR when `refresh = true` the existing `pulumi refresh && pulumi <command>` behavior is removed and commands are instead run with `--refresh`

Completely untested at this point.
I took a look at existing tests and didn't see an obvious place to add a test.
No local testing yet either, will play with that in a bit, erring on the side of opening early to enable discussion.


tgummerer:
@fitz-vivodyne Thanks for the PR and sorry about the delay here.  I've just let the tests run, and there are some failures.  I believe this has the same cause as https://github.com/pulumi/actions/pull/1116#pullrequestreview-1942880883, and I *think* the option needs to be added to the `action.yml` file.

fitz-vivodyne:
Thanks @tgummerer, just pushed the updates to `action.yml`

simenandre:
Hello 👋 Thanks for doing this, @fitz-vivodyne!

I am wondering what you think of keep using the `refresh` argument, and change the behavior of it to use this?

For most users they expect to be using this way of refreshing. We need to bump major, since it might break things, but it's the same kind of breaking we've done since v2; smaller and easy to migrate.

fitz-vivodyne:
>I am wondering what you think of keep using the refresh argument, and change the behavior of it to use this?
>For most users they expect to be using this way of refreshing. We need to bump major, since it might break things, but it's the same kind of breaking we've done since v2; smaller and easy to migrate.

@simenandre I found the current behavior surprising and when I first started using the action I 100% expected the `refresh` argument to just run with the `--refresh` flag.
If you actually want to run a `pulumi refresh` I would expect to run that as a standalone action/command.

Agreed that a major version bump would be appropriate, and I'm personally supportive of fully replacing the `refresh` functionality.

Thoughts on moving forward?
New PR, modify this one to strip out the refresh behavior?

komalali:
+1 to changing the refresh functionality and bumping the major version. I'm guessing the reason it behaved this way originally was because automation api didn't have support for the refresh flag. 

Although, now that I think about it, I think we've only added support for `--refresh` for [Go Automation API](https://github.com/pulumi/pulumi/pull/15350/files) so far, so we'll need to port that over to the nodejs automation api to be able to be used here.

mikhailshilkov:
@tgummerer I think we should provide the guidance for what we want to happen here. Let's make the decision and try to land this PR, or decide not to land it and describe the alternative.

tgummerer:
I'd be happy to land this PR if @fitz-vivodyne wants to update it with the updated `refresh` functionality.  Once that's done I can approve, merge and create a new release with a major version bump.

mikhailshilkov:
@fitz-vivodyne Will you have time to take another step on this PR?

fitz-vivodyne:
Sorry, haven't had bandwidth recently.
Might be a while before I'm able to get around to it, anyone else can feel free to grab it if they're available

mikhailshilkov:
@fitz-vivodyne Do you mind converting it to a Draft PR for now, so that it doesn't show up in our review queue?

fitz-vivodyne:
Alright, I find myself wanting this functionality again so I just pushed the discussed changes to the PR.

This now represents a breaking change/major release that would change the behavior of `refresh: true` to run commands with `--refresh` instead of running `pulumi refresh && pulumi <command>`

tgummerer (approved):
Neat, thank you!

mikocot:
@fitz-vivodyne @tgummerer as long as the effect of this change on `pulumi preview` has been explained and is clear to me' what is the change to `pulumi up` command? Is that just for consistency but we expect the same effect, with precison to maybe error-handling, or there is a significant difference in behavior too?

I believe this info should be included in the 'Migration from v5 section of the readme'. I think most people care about the actual behavior change than what is updated in the implementation.

PR #1146, opened 2024-04-11
Improve error message when version resolution fails

Frassle:
We've hit issues a few times where version resolution failed and it wasn't clear why because the error doesn't report the version returned from `pulumi version`. This adds that version (and the version we're trying for) to the error message to make debugging much easier.


komalali (approved):

justinvp:
Closing in favor of #1147

PR #1174, opened 2024-05-22
prepare for renaming master to main

tgummerer:
Prepare this repo for switching its main branch name to `main` by making the workflows run on both `master` and `main`.  Some workflows already run that way, but make sure all of them do, ahead of going ahead and renaming it.

Once the branch is renamed, we can drop the build on `master` here.


Frassle (approved):

justinvp:
FYI @simenandre

simenandre (approved):

PR #1178, opened 2024-05-24
make `comment-on-pr-number` option work even outside of pull-request events

### LINKED ISSUE #1177: comment-on-pr-number is ignored on push events
### What happened?

We upgraded our action version from 4.1.0 to 5.2.3 and noticed that we weren't getting any comment on PR when we merged.
Everything works correctly on pull_requests events but not anymore on push events and comment-on-number is set

### Example

```yaml
      - name: Retrieve current PR
        uses: jwalton/gh-find-current-pr@v1
        id: find_pr
        with:
          state: all

      - name: Run pulumi update
        uses: pulumi/actions@v5.2.3
        with:
          command: up
          stack-name: ${{ inputs.stack-name }}
          work-dir: ${{ inputs.work-dir }}
          cloud-url: '<redacted>'
          comment-on-pr-number: ${{ steps.find_pr.outputs.pr }}
          diff: true
          edit-pr-comment: false
```

### Output of `pulumi about`

CLI          
Version      3.116.0
Go Version   go1.22.3
Go Compiler  gc

Host     
OS       arch
Version  "rolling"
Arch     x86_64

Backend        
Name           ufmwk
URL            <redacted>
User           ulrich
Organizations  
Token type     personal

Pulumi locates its logs in /tmp by default

### Additional context

It sounds like the regression was introduced by this commit : https://github.com/pulumi/actions/commit/ca33bed63d58e10be6523f0184b5d368021a7d3d#diff-4fab5baaca5c14d2de62d8d2fceef376ddddcc8e9509d86cfa5643f51b89ce3dL114-R124

In fact, the code right now doesn't allow to comment on anything other than a pull_request event :

```typescript
  if (config.commentOnPrNumber || config.commentOnPr) {
    const isPullRequest = context.payload.pull_request !== undefined;
    if (isPullRequest) {  // handlePullRequestMessage isn't called if context.payload.pull_request is not set
      core.debug(`Commenting on pull request`);
      invariant(config.githubToken, 'github-token is missing.');
      handlePullRequestMessage(config, projectName, output);
    }
  }
```

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

tgummerer:
We only allow the `comment-on-pr` option inside of pull requests. This makes sense because we wouldn't be able to comment on anything otherwise.  However with the `comment-on-pr-number` option gives us a PR number to comment on, so it can make sense with any type of GitHub actions event.  This used to work before #813, and it looks like disallowing this was accidental rather than intentional.

Make this work again.

Fixes #1177

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## HEAD (Unreleased)
 
+- fix: `make comment-on-pr-number` option work even outside of pull-request events
+  ([#1178](https://github.com/pulumi/actions/pull/1178))
+
 ---
 
 ## 5.2.1 (2024-04-11)
--- a/src/main.ts
+++ b/src/main.ts
@@ -119,13 +119,12 @@ const runAction = async (config: Config): Promise<void> => {
     }
   }
 
-  if (config.commentOnPrNumber || config.commentOnPr) {
-    const isPullRequest = context.payload.pull_request !== undefined;
-    if (isPullRequest) {
-      core.debug(`Commenting on pull request`);
-      invariant(config.githubToken, 'github-token is missing.');
-      handlePullRequestMessage(config, projectName, output);
-    }
+  const isPullRequest = context.payload.pull_request !== undefined;
+  if (config.commentOnPrNumber ||
+      (config.commentOnPr && isPullRequest)) {
+    core.debug(`Commenting on pull request`);
+    invariant(config.githubToken, 'github-token is missing.');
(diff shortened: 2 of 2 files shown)


tgummerer:
The workflow failures here should be fixed by https://github.com/pulumi/actions/pull/1179.

julienp (approved):

PR #1204, opened 2024-06-24
feat: add option `pulumi-version-file`

corymhall:
Similar to how [Setup
Node](https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md?rgh-link-date=2024-06-24T19%3A12%3A42Z#node-version-file) action works, this adds a new `pulumi-version-file` option that allows users to specify a file to read the version from.

A lot of our providers use the `.pulumi.version` file to store the current version of pulumi and we want to be able to use that version rather than relying on what is already installed.

re https://github.com/pulumi/ci-mgmt/pull/990

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -23,8 +23,13 @@ export const installationConfig = rt.Record({
 export type InstallationConfig = rt.Static<typeof installationConfig>;
 
 export function makeInstallationConfig(): rt.Result<InstallationConfig> {
-  let pulumiVersion = getInput('pulumi-version') || '^3';
+  let pulumiVersion = getInput('pulumi-version');
   const versionFile = getInput('pulumi-version-file');
+  if (pulumiVersion && versionFile) {
+    throw new Error(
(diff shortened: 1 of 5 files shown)


danielrbradley (approved):
Love it!

src/config.ts
```diff
@@ -22,9 +23,20 @@ export const installationConfig = rt.Record({
 export type InstallationConfig = rt.Static<typeof installationConfig>;
 
 export function makeInstallationConfig(): rt.Result<InstallationConfig> {
+  let pulumiVersion = getInput('pulumi-version') || '^3';
```
tgummerer, line 26: Should we error out early if both `pulumi-version` and `pulumi-version-file` are given?

tgummerer (commented):
Do we want to document this option in the readme as well?

src/config.ts
corymhall: Good call! I'll update it.

tgummerer (approved):
Nice, thank you!

Issue #1215, opened 2024-07-01
`pulumi-version-file` only works in install-only mode


gustavkj:
### What happened?

It seems that the `pulumi-version-file` input argument only works in [install-only mode](https://github.com/pulumi/actions?tab=readme-ov-file#installation-only). I do not think this is expected behavior according to the documentation in the readme or in the PR: https://github.com/pulumi/actions/pull/1204

## When in non-install-only mode

When running the following action config:

```yml
      - name: Preview <project-name> changes
        uses: pulumi/actions@v5
        env:
          PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
        with:
          command: preview
          work-dir: ./projects/<project>
          comment-on-pr: true
          stack-name: prod
          pulumi-version-file: ./.pulumi.version
          cloud-url: ${{ steps.backend-url.outputs.result }}
```

I get this output:

```
> Run pulumi/actions@v5
  
Configured range: 
Matched version: v3.121.0
Install destination is /home/ubuntu/.pulumi
Successfully deleted pre-existing /home/ubuntu/.pulumi/bin
/usr/bin/tar xz --warning=no-unknown-keyword --overwrite -C /home/ubuntu/.pulumi -f /opt/actions-runner/_work/_temp/6800d063-e7de-41b2-86b5-d307745b54ce
Logging into s3://<i-removed-this>
pulumi preview on prod
```

Just to test, the content of the `.pulumi.version` file is:

```
3.120.0
```

## When in install-only mode

```yml
      - name: Install Pulumi
        uses: pulumi/actions@v5
        with:
          pulumi-version-file: ./.pulumi.version
      - name: Preview <project-name> changes
        uses: pulumi/actions@v5
        env:
          PULUMI_CONFIG_PASSPHRASE: ${{ secrets.PULUMI_CONFIG_PASSPHRASE }}
        with:
          command: preview
          work-dir: ./projects/<project>
          comment-on-pr: true
          stack-name: prod
          pulumi-version-file: ./.pulumi.version
          cloud-url: ${{ steps.backend-url.outputs.result }}
```

Output:

```
# Install Pulumi
> Run pulumi/actions@v5
Configured range: 3.120.0
Matched version: v3.120.0
Install destination is /home/ubuntu/.pulumi
Successfully deleted pre-existing /home/ubuntu/.pulumi/bin
[...]

PR #1218, opened 2024-07-03
make pulumi-version-file also work in non-install mode

tgummerer:
The pulumi-version-file option was introduced in https://github.com/pulumi/actions/pull/1204.  However it was only introduced for installation mode, not if a pulumi command is given. The latter has separate config parsing code, so let's introduce the same option there as well.

Fixes: https://github.com/pulumi/actions/issues/1215

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -47,6 +47,22 @@ export function makeInstallationConfig(): rt.Result<InstallationConfig> {
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 export function makeConfig() {
+  let pulumiVersion = getInput('pulumi-version');
+  const versionFile = getInput('pulumi-version-file');
+  if (pulumiVersion && versionFile) {
+    throw new Error(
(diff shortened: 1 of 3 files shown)


corymhall (approved):

src/config.ts
```diff
@@ -24,6 +24,9 @@ export type InstallationConfig = rt.Static<typeof installationConfig>;
 
 export function makeInstallationConfig(): rt.Result<InstallationConfig> {
   let pulumiVersion = getInput('pulumi-version');
+  if (pulumiVersion === "undefined") {
```
justinvp, line 27: When will this be the string `"undefined"`?

tgummerer: This is an awkward side effect of the way we're testing this. When we're setting up the mocked config [here](https://github.com/pulumi/actions/blob/main/src/__tests__/config.test.ts#L27), the "undefined" gets turned into a string. Though maybe we should be using mocks instead for at least the version so we can avoid this :thinking:

justinvp (approved):

Issue #1221, opened 2024-07-10
Using latest for pulumi-version option does not properly match the current and latest versions.


j-fulbright:
### What happened?

![image](https://github.com/pulumi/actions/assets/10793868/ec0d531d-432f-4def-b0d6-6afb5b18bddd)

This does not affect functionality, but one less thing to have to download or deal with in the process

### Example

See above

### Output of `pulumi about`

n/a

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

tgummerer:
Thanks for opening the issue.  Currently when `pulumi-version: latest` is specified we'll always try to download the latest version of the CLI, regardless of what's installed.  

https://github.com/pulumi/actions/pull/1225 moves that around, so we get the latest version first, and if that matches the currently installed version we won't download it.

Note that in the meantime you can work around this by removing the `pulumi-version` directive.  That means you'll be using the version installed on the runner by default, and a newer version only gets installed if pulumi is not installed on the runner, or if we marked the version that is installed as a known bad version.

PR #1225, opened 2024-07-11
only install new pulumi version when there's a newer one available

tgummerer:
Users can pass 'latest' as the 'pulumi-version' configuration setting. This currently means that they always download the latest version of the CLI, regardless of whether that's already installed or not.

We can do better here and check if the latest version matches the one already installed on the runner, and in that case skip the download.

Fixes https://github.com/pulumi/actions/issues/1221

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## HEAD (Unreleased)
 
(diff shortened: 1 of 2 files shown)


lunaris (approved):

julienp (approved):
Nice!

PR #1231, opened 2024-07-18
Remove ansi control chars from PR comment message

jmrt47:
Pulumi action used with `color: always` output and `comment-on-pr: true` lead to ANSI control characters in the Pull Request comment. Reported in #1229 check for details.

Special thanks to @bennettp123 who had the idea, converting the output from ansi to html and add html to the comment. I picked up the idea propsed in #859.

Changes proposed in this PR:
- Convert output from ansi to html and add html as a comment
- Add config to specifiy the maximum length of the PR comment output

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,9 @@ The action can be configured with the following arguments:
   `true`. This is in an effort to reduce verbosity - if you want to have a
   comment per PR run, please ensure that you set this to `false`.
 
+- `comment-max-character` - (optional) Maximum number of characters of the Pulumi
+  action result to be added to the PR. Default value is 64000.
+
 - `expect-no-changes` - (optional) Return an error if any changes occur during
   this update.
 
--- a/README.md
+++ b/README.md
@@ -87,9 +87,6 @@ The action can be configured with the following arguments:
   `true`. This is in an effort to reduce verbosity - if you want to have a
   comment per PR run, please ensure that you set this to `false`.
 
-- `comment-max-character` - (optional) Maximum number of characters of the Pulumi
-  action result to be added to the PR. Default value is 64000.
-
 - `expect-no-changes` - (optional) Return an error if any changes occur during
   this update.
 
--- a/src/libs/__tests__/pr.test.ts
+++ b/src/libs/__tests__/pr.test.ts
@@ -56,6 +56,25 @@ describe('pr.ts', () => {
     });
   });
 
+  it('should convert ansi controll character to html and add to pull request message', async () => {
+    // @ts-ignore
+    gh.context = {
+      payload: {
+        pull_request: {
+          number: 123,
+        },
+      },
+    };
+
+    process.env.GITHUB_REPOSITORY = 'pulumi/actions';
+
+    await handlePullRequestMessage(defaultOptions, projectName, '\x1b[30mblack\x1b[37mwhite');
+    expect(createComment).toHaveBeenCalledWith({
+      body: '#### :tropical_drink: `preview` on myFirstProject/staging\n\n<details>\n<summary>Pulumi report</summary>\n\n<pre>\n<span style="color:#000">black<span style="color:#AAA">white</span></span>\n</pre>\n\n</details>',
+      issue_number: 123,
+    });
+  });
+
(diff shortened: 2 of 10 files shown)


src/libs/__tests__/pr.test.ts
```diff
@@ -50,7 +51,26 @@ describe('pr.ts', () => {
 
     await handlePullRequestMessage(defaultOptions, projectName, 'test');
     expect(createComment).toHaveBeenCalledWith({
-      body: '#### :tropical_drink: `preview` on myFirstProject/staging\n\n<details>\n<summary>Pulumi report</summary>\n\n```\ntest\n```\n\n</details>',
+      body: '#### :tropical_drink: `preview` on myFirstProject/staging\n\n<details>\n<summary>Pulumi report</summary>\n\n<pre>\ntest\n</pre>\n\n</details>',
+      issue_number: 123,
+    });
+  });
+
+  it('should convert ansi controll character to html and add to pull request message', async () => {
```
tgummerer, line 59:
```suggestion
  it('should convert ansi control character to html and add to pull request message', async () => {
```

src/libs/pr.ts
```diff
@@ -1,9 +1,41 @@
...
+  let trimmed = false;
+
+  let htmlBody: string = convert.toHtml(message);
+
+  // Check if htmlBody exceeds max characters
+  if (htmlBody.length > maxLength) {
+
+    // trim input message by number of exceeded characters
+    const dif: number = htmlBody.length - maxLength;
+    message = message.substring(0, message.length - dif);
```
tgummerer, line 29: Hrm trimming it this way doesn't seem quite correct when we have HTML here, as we're risking trimming it in the middle of an HTML tag.  Not quite sure of the top of my head what the best way to go about that is.

Maybe we could convert it to HTML, check how much longer than allowed the message is, trim the original message by that much and then re-convert it to HTML?  Though that might shorten it by too much.  Maybe sort of a binary search style would work best to see how much we can fit into the max length?

README.md
```diff
@@ -87,6 +87,9 @@ The action can be configured with the following arguments:
   `true`. This is in an effort to reduce verbosity - if you want to have a
   comment per PR run, please ensure that you set this to `false`.
 
+- `comment-max-character` - (optional) Maximum number of characters of the Pulumi
```
tgummerer, line 90: Curious what the value is of making this configurable?  Iiuc we have the maximum character limit because there is a maximum character limit for GitHub comments (which seems to be 65535 https://github.com/orgs/community/discussions/27190#discussioncomment-4803699).  

So at best we can allow someone to reduce the maximum character count, but we should error out if this setting is set above 64000.

tgummerer (commented):
Thanks for getting this started!  I left a few comments inline.

src/libs/pr.ts
jmrt47: Actually I thought I'm doing exactly what you proposed:
1. Converting the `message` to html `htmlBody`
2. In case this result html is to long I check on how much `dif`
3. The original plain `message` is than trimmed by the amount of characters and converted to html again.
This might result in a bit shorter html message, which was adaptable for me keeping the complexity in mind to find the exact maximum html length.

Could you please cross check again and let me know if I missed something in the code that leads to a behavior other than described here.

tgummerer: Oh doh, sorry.  I totally misread this :facepalm: 

What you have here looks good.

README.md
jmrt47: You are right, with the limit of GitHub comments in mind the only option would be to lower the comment length. Let me remove the configuration option again and put it as a const in the pr.ts file instead.

src/libs/pr.ts
```diff
@@ -46,14 +46,16 @@ export async function handlePullRequestMessage(
     command,
     stackName,
     editCommentOnPr,
-    commentMaxCharacter,
   } = config;
 
+  // GitHub limits comment characters to 65535, use lower max to keep buffer for variable values
```
tgummerer, line 51: :heart: thanks for documenting this!

PR #1238, opened 2024-07-25
remove unnecessary await

tgummerer:
`makeConfig` is not an async function, so the await here has no effect.  Remove it to clean this up.

This is not a necessary change at all, but I need to merge a change to test the config settings I've changed for re-enabling mandatory code reviews for this repo, while also not having the issues mentioned in https://github.com/pulumi/actions/issues/1148.


lunaris (approved):

lunaris:
I feel like this is something the linter (or even type checker) can/should pick up -- something we could/should make stricter?

PR #1252, opened 2024-08-14
Trim PR comments from the front.

devonmoss:
I believe that the resources summary is the most important component of the output provided by pulumi preview and should always be included in the pull request comment. When trimming from the back, long outputs will not contain this information, which requires navigating to the checks tab and looking at the output there. 

Perhaps this is just a reflection of how I use pulumi but don't think so because we are all conditioned by the local use of the pulumi cli. When we run pulumi preview locally the output ends up right at the bottom so the first thing we see is the summary something like this:
```
Resources:
      ~ 85 to update
      +-26 to replace
      111 changes. 1062 unchanged
```
In many cases this is all the information that is needed to confirm that I'm ready to run pulumi up. Other times I need to look more closely at the details for each resource but even still I have the information from that summary to give me an idea of what I'm looking at as I scroll back through the output.

This change trims the PR comment from the front if it is ever so long that it must be trimmed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,11 +143,16 @@ The action can be configured with the following arguments:
 
 - `plan` - (optional) Used for
   [update plans](https://www.pulumi.com/docs/concepts/update-plans/)
+
   - On `preview`: Where to save the update plan. If you choose to use this in a
     different run of your workflow (let's say you create an update plan via a
     preview on Pull Request creation and want to `up` using the plan) you must
     upload the plan as an artifact, and retrieve it wherever you run `up`
   - On `up`: Where to read the update plan from.
+
+- `trim-comments-from-front` - (optional) If `true`, then the action will trim
+  long pr comments from the front instead of the back. This ensures that the
+  resources summary is always visible.
 
 By default, this action will try to authenticate Pulumi with
 [Pulumi Cloud](https://app.pulumi.com/). If you have not specified a
--- a/README.md
+++ b/README.md
@@ -150,9 +150,9 @@ The action can be configured with the following arguments:
     upload the plan as an artifact, and retrieve it wherever you run `up`
   - On `up`: Where to read the update plan from.
 
-- `trim-comments-from-front` - (optional) If `true`, then the action will trim
+- `always-include-summary` - (optional) If `true`, then the action will trim
   long pr comments from the front instead of the back. This ensures that the
-  resources summary is always visible.
+  resources summary is always included in the comment.
 
 By default, this action will try to authenticate Pulumi with
 [Pulumi Cloud](https://app.pulumi.com/). If you have not specified a
--- a/src/libs/__tests__/pr.test.ts
+++ b/src/libs/__tests__/pr.test.ts
@@ -144,7 +144,7 @@ describe('pr.ts', () => {
 
     const call = createComment.mock.calls[0][0];
     expect(call.body.length).toBeLessThan(65_536);
-    expect(call.body).toContain('The output was too long and trimmed');
+    expect(call.body).toContain('The output was too long and trimmed.');
   });
 
   it('should trim the output from front when the output is larger than 64k characters and config is set', async () => {
@@ -173,7 +173,7 @@ describe('pr.ts', () => {
     const call = createComment.mock.calls[0][0];
     expect(call.body.length).toBeLessThan(65_536);
     expect(call.body).toContain('this is at the end and should be in the output');
-    expect(call.body).toContain('The output was too long and trimmed from the front');
+    expect(call.body).toContain('The output was too long and trimmed from the front.');
   });
 
   it('should edit the comment if it finds a previous created one', async () => {
--- a/src/libs/__tests__/pr.test.ts
+++ b/src/libs/__tests__/pr.test.ts
@@ -158,15 +158,15 @@ describe('pr.ts', () => {
         },
       },
     };
-    const trimFromFrontOptions = {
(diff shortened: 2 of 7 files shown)


tgummerer:
Thanks for the contribution.  I'm not sure I agree that this is better than what we have currently though.  Could you give us a bit of a description of why you think this is better?  (This should also be included in the PR description) Maybe we could make this behaviour optional and add another flag to toggle it?  Trimming from the front of the message seems unconventional and should probably not be the default behaviour.

devonmoss:
I've updated the description for my PR. 

I agree that generally speaking, trimming from the front is unconventional but in this case it causes a scenario where the resources summary is not shown in the pull request comment and I can't see why this would ever be a desirable situation.

voldemortensen:
I agree with @devonmoss here. When I'm reviewing potential changes, I am much less interested in the output of the pip install and much more interested in which resources are going to be effected by a merge request. In fact, unless there's an error with pip install, I'm not really interested in that output at all.

devonmoss:
I've added a commit which allows this to be configured optionally. My apologies for not reading the CONTRIBUTING.md file before submitting. I see now that I should have submitted an issue first. I'll be better next time.

README.md
```diff
@@ -143,12 +143,17 @@ The action can be configured with the following arguments:
...
 - `plan` - (optional) Used for
   [update plans](https://www.pulumi.com/docs/concepts/update-plans/)
+
   - On `preview`: Where to save the update plan. If you choose to use this in a
     different run of your workflow (let's say you create an update plan via a
     preview on Pull Request creation and want to `up` using the plan) you must
     upload the plan as an artifact, and retrieve it wherever you run `up`
   - On `up`: Where to read the update plan from.
 
+- `trim-comments-from-front` - (optional) If `true`, then the action will trim
```
tgummerer, line 153:
```suggestion
- `always-include-summary` - (optional) If `true`, then the action will trim
```

Trimming from the front is what it does, but maybe this makes the intention clearer?

src/libs/__tests__/pr.test.ts
```diff
@@ -161,7 +192,7 @@ describe('pr.ts', () => {
     await handlePullRequestMessage(options, projectName, 'test');
     expect(updateComment).toHaveBeenCalledWith({
       comment_id: 2,
-      body: '#### :tropical_drink: `preview` on myFirstProject/staging\n\n<details>\n<summary>Pulumi report</summary>\n\n<pre>\ntest\n</pre>\n\n</details>',
+      body: '#### :tropical_drink: `preview` on myFirstProject/staging\n\n<details>\n<summary>Pulumi report</summary>\n\n\n<pre>\ntest\n</pre>\n\n</details>',
```
tgummerer, line 195: Why are we getting this extra newline everywhere?  I don't think we want that :thinking:

tgummerer (commented):
> I've added a commit which allows this to be configured optionally. 

Thanks, this is great!  I think we should be able to flip this to be the default in the next major release, still giving the users the option to opt-out.  I'm mainly wary of changing this in a minor release.

> My apologies for not reading the CONTRIBUTING.md file before submitting. I see now that I should have submitted an issue first. I'll be better next time.

No worries.  It helps sometimes to discuss the details first to avoid some roundtrips in PRs, but we're always happy to review PRs.  I left a couple of minor comments inline, once they are addressed I think we can get this merged!

README.md
devonmoss: Yeah, I like that.

src/libs/pr.ts
```diff
@@ -55,21 +64,24 @@ export async function handlePullRequestMessage(
...
-  const [htmlBody, trimmed]: [string, boolean] = ansiToHtml(output, MAX_CHARACTER_COMMENT);
+  const [htmlBody, trimmed]: [string, boolean] = ansiToHtml(output, MAX_CHARACTER_COMMENT, trimCommentsFromFront);
 
   const body = dedent`
     ${heading}
 
     <details>
     ${summary}
 
+    ${trimCommentsFromFront
```
devonmoss, line 75: This is the reason why I added an extra new line. Previously, when the output was trimmed it would add a warning message to the end of the comment. My change makes it so in the event that we trim from the front it will add the warning message near the beginning of the comment, right before the trimmed output.

The blank line above was not there during the first pass I made and the tests passed without the addition of the extra `\n` to the tests. However when I used it in a project the markdown doesn't format properly. Instead of:

:warning: **Warn**: The output was too long and trimmed from the front.

It did:

:warning: \*\*Warn\*\*: The output was too long and trimmed from the front.

src/libs/__tests__/pr.test.ts
devonmoss: https://github.com/pulumi/actions/pull/1252/files#r1725676206

I'll test this again and take a couple of screenshots so you can see exactly what I mean. I didn't notice any gratuitous spacing but I didn't look for it explicitly.

devonmoss: I think it has something to do with coming right after the `</summary>` tag. The issue can be recreated in any comment.
<summary>This summary is good</summary>

⚠️ **Woohoo** look I'm bold

<summary>This is bad</summary>
⚠️ **Woohoo** look I'm bold?

devonmoss: I think the extra `\n` needs to be in the tests

tgummerer (approved):
Thanks for implementing this!

PR #1263, opened 2024-08-22
only log warning if output is actually trimmed

devonmoss:
Fixing small issue where the warning message would always be added to the commit message if `always-include-summary` was set to `true`. This fixes it so it only shows when the output is trimmed.


tgummerer (approved):
Thanks!

PR #1299, opened 2024-10-22
Make `pulumi login` respect configuration in `Pulumi.yaml`

### LINKED ISSUE #1120: Make `pulumi login` respect configuration in `Pulumi.yaml`
While working on https://github.com/pulumi/actions/issues/1081, we noticed that `pulumi login` wasn't respecting the configuration in `Pulumi.yaml`, and thus errored out when no `cloud-url` was given in the action.  We should improve `pulumi login`, so it takes the setting in `Pulumi.yaml` into account if that file is available to reduce confusion for users.

/xref https://github.com/pulumi/home/issues/3298

tlinhart:
Fixes #1120

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## HEAD (Unreleased)
 
+- fix: Make `pulumi login` respect configuration in `Pulumi.yaml`
+  ([#1299](https://github.com/pulumi/actions/pull/1299))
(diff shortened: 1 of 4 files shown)


tgummerer (approved):
Sorry about the delay here @tlinhart, and thanks so much for the contribution!  This seems like the right thing to do indeed!

PR #1300, opened 2024-10-24
add workflow for publishing immutable actions

tgummerer:
We're switching over to publishing immutable actions from this repo. For this we need to make sure that immutable actions are published at each release of this repo.

With this, we can keep following our current release process, and whenever a new release is published, this action will create a new immutable actions package, that will then be available to users.

(This should only be merged after the current migration process (see https://github.com/pulumi/actions/actions/workflows/immutable-actions-migration/migrate_release) is completed, so that packages will always be in the right order.  Though it wouldn't be a big deal if they were out of order either.  Cosmetically it's nicer if they are in order though)

Fixes https://github.com/pulumi/home/issues/3695

### DIFF
--- a/.github/workflows/immutable-action.yml
+++ b/.github/workflows/immutable-action.yml
@@ -1,0 +1,23 @@
+name: "Publish Immutable Action Version"
+
+on:
+  release:
+    types: [created]
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+      packages: write
+
+    steps:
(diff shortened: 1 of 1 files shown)


julienp (approved):
LGTM, assuming that the action automatically takes the release that trigged it

tgummerer:
> LGTM, assuming that the action automatically takes the release that trigged it

Yup, if you specify `pulumi/actions@v6` it will take the latest semver release for v6 (something on the server side does this for mutable actions).  You can also specify specific full tags directly (or git OIDs) directly.

PR #1312, opened 2024-11-25
Remove ANSI control chars from GitHub step Summary and PR comment

### LINKED ISSUE #1248: Colors code on summary should not be printed
## Issue details

Related to #1228. I'm seeing color-related characters in the GitHub step summary when `color: always` and `comment-on-summary: true` are both set.

The action allows to enable `color` with `auto` (default), `raw`, `never` and `always`.
The action allows to show a `comment-on-summary` for an update.

It's not possible to enable colors only for the GHA output:

1) enable `color: always`
1) the preview/output in the "Actions" github page is shown with colors
1) the output in the summary of the check contains color and escape codes

Should be fair to allow colors only for the "Actions" pulumi output and not in the comment of the PR.

### Affected area/feature

Pulumi output/colors

byian:
This PR introduces striping of ANSI symbols for GitHub step Summary and improves the current one for PR comments.

To resolve #1248

This is similar to #1231, but uses a slightly different approach. Instead of converting ANSI symbols to html tags (using the `ansi-to-html` package), which results in `\x1b[30mblack` becoming `<span style="color:#000">black</span>`, we simply remove them using a regular expression, which results in `\x1b[30mblack` becoming just `black`. 

The reason is that neither the Summary nor the PR comment is able to actually render the `<span style="color:#000">` block with color. However, all these html tags consume limits for PR comment length and Summary size.

### DIFF
--- a/src/libs/summary.ts
+++ b/src/libs/summary.ts
@@ -65,11 +65,13 @@ export async function handleSummaryMessage(
   }
 
   const body = dedent`
+    <pre lang="diff"><code>
     ${message}
-  `
+    </code></pre>
+  `;
 
   await core.summary
-  .addHeading(heading)
-  .addCodeBlock(body, "diff")
-  .write();
+    .addHeading(heading)
+    .addRaw(body)
+    .write();
 }
--- a/src/libs/summary.ts
+++ b/src/libs/summary.ts
@@ -48,8 +48,11 @@ export async function handleSummaryMessage(
   } = config;
 
   // strip ANSI symbols from message because it is not supported in GH step Summary
-  const regex = RegExp(`\x1B(?:[@-Z\\-_]|[[0-?]*[ -/]*[@-~])`, 'g');
-  output = output.replace(regex, '');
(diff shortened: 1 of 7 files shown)


byian:
Hi @tgummerer, do you by any chance have time to take a look at this PR?

byian:
Hi all
I updated this PR, and I'm still waiting for review. 

This one contains some real QoL improvements

src/libs/summary.ts
```diff
@@ -0,0 +1,74 @@
+import * as core from '@actions/core';
+import { Config } from '../config';
+
+function trimOutput(
```
Frassle, line 4: Why trimOutput here and in pr.ts? Looks like one of these should be removed they're doing the same thing.

byian: Basically, they do the same thing, yes, they trim the input string to the desired format, but for pr and summary, the formats are different. 
Summary has a limit on the maximum message size in bytes, and pr has a limit on the maximum length in characters.
Maybe I should use the different names to avoid confusion?

PR #1327, opened 2024-12-23
Fix the `output` command

### LINKED ISSUE #868: Support for 'stack output' command

justinvp:
This commit fixes the `output` command to work without having to have a workDir specified to a location with a Pulumi.yaml file. It does this by avoiding calls to `LocalWorkspace.createOrSelectStack`/`selectStack` when the command is `output`. This allows the `output` command to work with any stack, even when there is not a local project available.

Additionally, when the command is `output`, don't attempt to comment on the PR; only actual pulumi operations (like the `up` command) should do that.

Fixes #868

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,6 +5,8 @@ import {
   LocalProgramArgs,
   LocalWorkspace,
   LocalWorkspaceOptions,
+  OutputMap,
+  Stack,
 } from '@pulumi/pulumi/automation';
 import invariant from 'ts-invariant';
 import {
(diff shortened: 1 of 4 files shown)


src/main.ts
```diff
@@ -54,29 +56,42 @@ const runAction = async (config: Config): Promise<void> => {
...
-    stackOpts.secretsProvider = config.secretsProvider;
+    wsOpts.secretsProvider = config.secretsProvider;
   }
 
-  const stack = await (config.upsert
-    ? LocalWorkspace.createOrSelectStack(stackArgs, stackOpts)
-    : LocalWorkspace.selectStack(stackArgs, stackOpts));
+  // Only initialize `stack` when the command is not `output`.
+  // When the command is `output` we want to avoid the underlying call to `pulumi stack select`,
+  // which requires a Pulumi.yaml file to be present.
```
justinvp, line 66: The only downside is that we aren't respecting `config.upsert` for the `output` command, so if the stack doesn't exist, it won't be created. I actually think this is ok and the behavior we want for `output`. You really only want to get stack outputs for a stack that exists.

If we felt it was important to keep `upsert` working for the `output` command, we could make a change below on line 131: when calling `outputs = await ws.stackOutputs(config.stackName);` we could wrap that in a try/catch and if it's a `StackNotFoundError` and `config.upsert` is true, we could create the stack and return empty outputs. My preference is to not do this unless we get feedback that it's needed for some scenario.

Zaid-Ajaj (approved):

darena-patrick:
very excited to try this out - is there a version we should target to provide feedback?

PR #1352, opened 2025-02-05
feat: Add working-directory alias for work-dir input

n0mer:
Add working-directory as an alias for work-dir input to improve compatibility with other GitHub Actions.


action.yml
```diff
@@ -21,6 +21,10 @@ inputs:
     description: 'Location of your Pulumi files. Defaults to ./'
     required: false
     default: ./
```
Frassle, line 23: Don't we need to remove this default otherwise getInput will always get this default value?

simenandre: It seems so, yeah.

It might be better to remove the default on `work-dir`.

Following that change, we can deprecate `work-dir` but keep it as an alias to `working-directory` for backwards compatibility.

src/__tests__/config.test.ts
```diff
@@ -199,4 +199,76 @@ describe('config.ts', () => {
...
+      getMultilineInput: jest.fn((name: string) => {
+        return defaultConfig[name];
+      }),
+    }));
+    const { makeConfig } = require('../config');
+    const c = makeConfig();
+    expect(c.workDir).toBe('/custom/path');
+  });
+
+  it('should prefer work-dir over working-directory when both are provided', async () => {
```
simenandre, line 251: I think it's more transparent for the consumers if the user cannot use both. For example, we can throw an error if the user supplies both.


What do you think?

PR #1355, opened 2025-02-26
Add a clickable View Live link before pre-block

Mctalian:
👋🏼 

I brought this up in my company's shared Slack with y'all. Seemed pretty low effort so I decided to forego the issue creation.

The idea is to offer a clickable link to the update run. Currently, since it is within the `<pre>` tags, the link is not clickable and must be highlighted in a window that is likely horizontally scrolling. The other alternative is to find the clickable link within the GitHub Action logs.

We've seen a few times where both the author and the reviewer have failed to double-check the preview run to ensure we're not causing any surprises. My hope is that by making it easier to double-check the preview output in Pulumi, that the likelihood of forgetting to check the output will decrease.

I could see an argument for putting the `View Live` link outside of the `<details>` section entirely so it is only a click away, but I wanted to propose these changes as they are to get a discussion started.

Thanks!

### DIFF
--- a/src/libs/pr.ts
+++ b/src/libs/pr.ts
@@ -88,7 +88,7 @@ export async function handlePullRequestMessage(
 
     <details>
     ${summary}
-    ${viewLiveLink ? `[View Live](${viewLiveLink})\n` : ''}
+    ${viewLiveLink ? `[View in Pulumi Cloud](${viewLiveLink})\n` : ''}
     ${trimmed && alwaysIncludeSummary
       ? ':warning: **Warn**: The output was too long and trimmed from the front.'
       : ''
--- a/src/libs/pr.ts
+++ b/src/libs/pr.ts
@@ -88,7 +88,7 @@ export async function handlePullRequestMessage(
 
     <details>
     ${summary}
-    ${viewLiveLink ? `[View in Pulumi Cloud](${viewLiveLink})\n` : ''}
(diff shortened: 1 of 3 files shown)


src/libs/pr.ts
```diff
@@ -66,12 +81,14 @@ export async function handlePullRequestMessage(
...
 
+  const viewLiveLink = extractViewLiveLink(output);
+
   const body = dedent`
     ${heading}
 
     <details>
     ${summary}
-
+    ${viewLiveLink ? `[View Live](${viewLiveLink})\n` : ''}
```
tgummerer, line 91: Very minor bikeshedding here, but maybe 
```suggestion
    ${viewLiveLink ? `[View in Pulumi Cloud](${viewLiveLink})\n` : ''}
```
would be slightly clearer? When the comment is posted the update has already finished, so there's no viewing it "live" anymore.

tgummerer (commented):
Thanks for the contribution! This seems useful to me.

I got a very small bikesheddy comment inline. Let me know what you think and then we can get this merged and released.

src/libs/pr.ts
Mctalian: Yes I agree, that's better. I'll update things in a bit!

tgummerer (approved):
Awesome, thanks!

PR #1357, opened 2025-03-05
feat: support continue on error flag

### LINKED ISSUE #1306: Continue on error as an option
## Hello!

## Issue details

Hi! 👋 

It would be awesome to leverage `--continue-on-error` ([link to Pulumi docs](https://www.pulumi.com/blog/continue-on-error/)) in the Pulumi github action! This is a very useful feature for when we have parts of the stack that are separate from each other, and we don't want to stop the Pulumi up command due to an unrelated error. 

Thanks for all the hard work 🙏 

### Affected area/feature

Not sure!

UnderKoen:
Adds support for the `--continue-on-error` flag

Resolves #1306

### DIFF
--- a/src/__tests__/config.test.ts
+++ b/src/__tests__/config.test.ts
@@ -52,6 +52,7 @@ describe('config.ts', () => {
         "githubToken": "n/a",
         "options": Object {
           "color": undefined,
+          "continueOnError": false,
           "diff": false,
           "excludeProtected": false,
           "expectNoChanges": false,


tgummerer:
Awesome, thanks!  I think `continueOnError` needs to be added below [here](https://github.com/pulumi/actions/pull/1357/files#diff-9ec2cf1ec8ef54c60ab38268ea30307b18d817bf6f95b57db312c655b8426480R66) as well.   I've started the tests, so they will probably tell us as well.

tgummerer (approved):
Thanks for the contribution, this is great!

PR #1363, opened 2025-04-25
Expose `--exclude` and `--exclude-dependents`

### LINKED ISSUE #1362: Support --exclude flags
## Hello!

## Issue details

Hello! I'd like to be able to leverage the [newly minted](https://github.com/pulumi/pulumi/pull/18828) `--exclude` flag for the commands that support it inside the github action.

My use case specifically:
We need it for refresh, or `preview --refresh` ideally we'd be able to ignore a urn on the "refresh" part of a `preview` but still get a preview for it, if possible. Though I would also be fine just splitting the steps out for explicit excludes.

Example
```yaml
      - name: Pulumi Refresh
        uses: pulumi/actions@v6
        with:
          command: refresh
          exclude: 'urn:...`
```

Elaboration
```yaml
      - name: Pulumi Refresh
        uses: pulumi/actions@v6
        with:
          command: preview
          refresh: true
          exclude: 'urn:...` # <== is exclude applied to both refresh/preview? i assume it would be.
```

### Affected area/feature

`pulumi/actions@v6` specifically `refresh` or `preview --refresh` commands.

i-am-tom:
This PR adds `--exclude` and `--exclude-dependents` to the configuration object for the GitHub action, which will then pass them through to the Automation API underneath.

Fixes #1362.


tgummerer (commented):
Looks good, but you don't need to update `dist/*` or `yarn.lock`.  The tooling will do that automatically after the PR is merged (it might be fine to do it here, but I vaguely remember the tooling getting slightly confused when I did it, so it's probably better to revert these changes).

We should write this down somewhere if we haven't already.

i-am-tom:
I'm confused about `yarn.lock`; why do we not check in the version we used to make the change?

tgummerer:
Hrm I might have been wrong about the `yarn.lock`, sorry.  https://github.com/pulumi/actions/blob/main/.github/workflows/update_dist.yml is what updates it, but it only updates files in `dist/`.

tgummerer (approved):
Sorry this slipped under my radar.  Looks good to me!

PR #1365, opened 2025-04-29
Use ESC secrets

pgavlin:
These changes migrate this repo's GitHub Actions Workflows to use ESC secrets instead of GitHub Secrets.

The changes are largely mechanical:

- Common configuration for all ESC actions within a workflow is added to the workflow's environment variables
- Permissions are expanded as necessary for workflows that do not grant `id-token: write` permissions
	- `read-all` permissions are replaced with the union of all explicit read permissions and `id-token: write`
	- Default permissions are replaced with `write-all`, which is the equivalent of all explicit write permissions and
	  `id-token: write`
	- Explicit permissions are modified to grant `id-token: write`
- A step that fetches ESC secrets and populates environment variables is added to each step that reads secrets
- Direct references to secrets within the job are replaced with references to the step's outputs

All ESC actions are configured to fetch secrets from a shared ESC environment that contains secrets migrated from GitHub Actions. The ESC action performs its own OIDC exchange to obtain a Pulumi Access Token.

### DIFF
--- a/.github/workflows/update_dist.yml
+++ b/.github/workflows/update_dist.yml
@@ -5,15 +5,22 @@ on:
       - main
 
 env:
-  GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+  ESC_ACTION_OIDC_AUTH: true
+  ESC_ACTION_OIDC_ORGANIZATION: pulumi
+  ESC_ACTION_OIDC_REQUESTED_TOKEN_TYPE: urn:pulumi:token-type:access_token:organization
+  ESC_ACTION_ENVIRONMENT: imports/github-secrets
+  ESC_ACTION_EXPORT_ENVIRONMENT_VARIABLES: GITHUB_TOKEN=PULUMI_BOT_TOKEN
 
 jobs:
   update-dist:
     runs-on: ubuntu-latest
     steps:
+      - name: Fetch secrets from ESC
+        id: esc-secrets
+        uses: pulumi/esc-action@v1
       - uses: actions/checkout@v4
         with:
-          token: ${{ secrets.PULUMI_BOT_TOKEN }}
+          token: ${{ steps.esc-secrets.outputs.PULUMI_BOT_TOKEN }}
       - uses: actions/setup-node@v4
         with:
           node-version: 20.x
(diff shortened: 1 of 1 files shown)


.github/workflows/update_dist.yml
```diff
@@ -1,19 +1,28 @@
+permissions:
+  id-token: write
```
tgummerer, line 2: Just to make sure I understand this right, this is so ESC can authenticate using `OIDC`?  

Do we not need the `auth-actions` as described here? https://github.com/pulumi/esc-action?tab=readme-ov-file#full-example-using-pulumiauth-actions-for-authentication

It feels like some of this could benefit from better documentation for the action.

.github/workflows/update_dist.yml
```diff
@@ -1,19 +1,28 @@
...
     branches:
       - main
 
 env:
-  GITHUB_TOKEN: ${{ secrets.PULUMI_BOT_TOKEN }}
+  ESC_ACTION_OIDC_AUTH: true
+  ESC_ACTION_OIDC_ORGANIZATION: pulumi
+  ESC_ACTION_OIDC_REQUESTED_TOKEN_TYPE: urn:pulumi:token-type:access_token:organization
+  ESC_ACTION_ENVIRONMENT: imports/github-secrets
+  ESC_ACTION_EXPORT_ENVIRONMENT_VARIABLES: GITHUB_TOKEN=PULUMI_BOT_TOKEN
```
tgummerer, line 14: Where can I find documentation for these environment variables?  Google doesn't seem to find anything, and I can't find anything on the `esc-actions` readme either.

.github/workflows/update_dist.yml
pgavlin: Yes, definitely. Will get that updated.

.github/workflows/update_dist.yml
pgavlin: Commit history :/

I'll update the README

.github/workflows/update_dist.yml
```diff
@@ -27,3 +34,5 @@ jobs:
           commit_message: 'chore(dist): Update dist [automated commit]'
           commit_user_name: Pulumi Bot
           commit_user_email: pulumi-bot@users.noreply.github.com
+permissions:
+permissions: write-all # Equivalent to default permissions plus id-token: write
```
komalali, line 38:
```suggestion
permissions: write-all # Equivalent to default permissions plus id-token: write
```

komalali (approved):

.github/workflows/update_dist.yml
pgavlin: Done. The README for the action now documents all of this.

komalali (approved):

PR #1367, opened 2025-05-07
update "user agent" string from `pulumi/actions@v5` to `pulumi/actions@v6`

EronWright:
The action code provides a user agent string to the automation api, for informational purposes.  Looks like the string is out of date; it still says "v5".  For example, some output from an unrelated issue:

```
stderr: Command failed with exit code 255: pulumi up --yes --skip-preview --diff --exec-agent pulumi/actions@v5 --color auto --exec-kind auto.local --stack dev-eus-00-wil-in1 --non-interactive
[resource plugin random-4.18.2] installing
[resource plugin random-4.18.1] installing
[resource plugin random-4.18.0] installing
[resource plugin azuread-6.4.0] installing
[resource plugin azure-native-3.2.0] installing
error: Status=400 Code="CertificateInUse" Message="Certificate 'cert-d5add54c6ad5054e50af24f769f45cf5' is used by existing custom domains."
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@
 ## HEAD (Unreleased)
 
 - feat: verbosity and debug inputs to enable debug options
+  ([#1046](https://github.com/pulumi/actions/pull/1046))
+
+- fix: update "user agent" string to pulumi/actions@v6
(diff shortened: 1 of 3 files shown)


tgummerer (approved):
Ah yes, we keep forgetting to update this.  Thanks for fixing!

PR #1370, opened 2025-05-12
Add workflow to export repo secrets

pgavlin:
This repository has repository-specific secrets that need to be migrated to ESC. These changes add a GitHub Actions workflow to perform this migration.

The workflow is configured to be run manually. It is only expected that the worflow will be run once. The workflow uses https://github.com/pulumi/esc-export-secrets-action to create a new ESC environment to hold the repository's secrets. This environment also imports the environment that holds the organization's secrets.

### DIFF
--- a/.github/workflows/export-repo-secrets.yml
+++ b/.github/workflows/export-repo-secrets.yml
@@ -1,0 +1,25 @@
+permissions: write-all # Equivalent to default permissions plus id-token: write
+name: Export secrets to ESC
+on: [ workflow_dispatch ]
+jobs:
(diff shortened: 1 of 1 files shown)


komalali (approved):

justinvp:
What are the repo-specific secrets? I only see `ACTIONS_RUNNER_DEBUG` and `ACTIONS_STEP_DEBUG` and I'm pretty sure those are just set to `true` per https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/troubleshooting-workflows/enabling-debug-logging

Issue #1364, opened 2025-04-29
Pulumi CLI does support windows on arm64


danielrbradley:
This failure is incorrect and blocks usage of the action on windows-arm64:

https://github.com/pulumi/actions/blob/9519177da243fd32cab35cdbf19cce1ab7472fcc/src/libs/pulumi-cli.ts#L65

The [latest Pulumi CLI release](https://github.com/pulumi/pulumi/releases/tag/v3.165.0) does include the binary [pulumi-v3.165.0-windows-arm64.zip](https://github.com/pulumi/pulumi/releases/download/v3.165.0/pulumi-v3.165.0-windows-arm64.zip).

Example failure: https://github.com/pulumi/verify-provider-release/actions/runs/14733350173/job/41353055483?pr=144

mjeffryes:
fixed by https://github.com/pulumi/actions/pull/1384[](url)

PR #1384, opened 2025-08-11
Adding ARM support for win32 

Graham-Pedersen:
https://github.com/pulumi/actions/issues/1364

We want to support ARM on windows. Currently this check fails even though we have much of infrastructure to support ARM.


Graham-Pedersen:
Fixing, the test suite currently, sorry for the churn.

justinvp (approved):
LGTM

PR #1385, opened 2025-08-20
Add preview-only flag support for destroy command

### LINKED ISSUE #1275: Pulumi Destroy support for --preview-only option
## Allow --preview-only option for destroy command!

## Issue details

Pulumi Destroy CLI command allows a --preview-only option to be provided. This github action doesnt seem to support preview-only destroy commands. Can this functionality be added?

scjody:
Closes #1275 

**Note that I vibe coded this and haven't yet reviewed or tested it. I'll mark the PR as ready for review and remove this message once it's ready!**

This adds a new `preview-only` input parameter that allows users to preview what would be destroyed without actually performing the destruction. This is equivalent to running `pulumi destroy --preview-only` from the CLI.

Changes:
- Added `preview-only` input parameter to action.yml
- Updated config.ts to handle the new parameter
- Added tests for the new functionality
- Updated README.md documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,9 @@ The action can be configured with the following arguments:
 - `exclude-protected` - (optional) Skip destroying protected resources. Only
   valid when `command` is `destroy`.
 
+- `preview-only` - (optional) Only show a preview of the operation, but do not
+  perform it. Only valid when `command` is `destroy`.
+
 - `suppress-outputs` - (optional) Suppress display of stack outputs (in case
   they contain sensitive values).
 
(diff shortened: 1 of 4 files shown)


tgummerer:
Just a heads up, this is not going to be as simple as it looks at first glance.  GitHub Actions uses the pulumi automation api under the hood. `destroy` does support a "preview only" flag, but I believe it's deprecated because it does not work correctly. 

Although on second look, we don't even seem to have implemented `previewDestroy` in the NodeJS automation API yet.  @i-am-tom https://github.com/pulumi/pulumi/pull/19948 seems to hint that we should implement `previewDestroy` for the NodeJS automation API as well, but it doesn't look like we ever did that.  Was that an oversight, or does the `preview_only` flag work for destroy in the automation API?

PR #1391, opened 2025-10-03
feat: Add option to omit stdout from the action's outputs

ionicsolutions:
Adds an option to exclude the captured stdout in the action's output.

This can be useful when using the action's output to populate the environment of  subsequent workflow steps by setting `env: ${{ steps.pulumi.outputs }}` or jointly exporting the stack outputs to other steps and jobs via `toJSON(steps.pulumi.outputs)`, where they can then be retrieved as, e.g., `env: ${{ fromJSON(needs.pulumi.outputs.stack-outputs) }}`.


ionicsolutions:
> We would like to add a new top level output key: `stack-outputs`, as you suggested in your original PR description. To preserve backward compatibility, each stack output would be set at both the top level and nested inside `stack-outputs`.

I can look into this, but it's not a high priority for me right now.

iwahbe:
Closing for now. If you are willing to make the changes discussed, we would be happy to receive another PR.

PR #1401, opened 2026-03-02
feat: Use diff syntax highlighting for PR comments when color is always

simone-panico:
## Summary                                                                                                                                                                                    
  - When `color: always` is set, PR comments use `diff` code blocks instead of `<pre>` tags, colorizing lines with `+`, `-`, and `~` markers                                                   
  - Diff markers are moved to line start so GitHub can apply syntax highlighting

## Example

Screenshot 2026-03-02 at 13 37 25


peyerc (approved):

Frassle:
Hi @simone-panico and thanks for the idea, but while this is a neat trick it doesn't feel "correct" to the team. We don't guarantee the +/- characters it's just incidental they line up with diff here.

PR #1405, opened 2026-04-29
Bump TypeScript and jest toolchain

nelsonfigueroa:
Bumps TypeScript and jest toolchain to more modern versions.

I ran `yarn install`, `yarn test`, and `yarn build` locally with no issues.


iwahbe (approved):
This looks fine. Since this isn't a library, I don't think there is any reason not to do this.

PR #1411, opened 2026-06-04
Add 6.5.0 release to changelog

leejones:
I thought the `run-program` option was supported because it was added in `6.4.0`, but later discovered it was removed in `6.5.0`. This PR includes the removal in the changelog to make that more visible.


Frassle (approved):
Thanks, not sure why this wasn't in the changelog to start.

PR #1434, opened 2026-06-22
Bump jest-playback to v4 and move get-version off `got`

iwahbe:
Follow-up to #1429, which deferred the `jest-playback` 3 → 4 bump because v4 is ESM-only and a ground-up rewrite.

## What changed

- **Bump `jest-playback` to ^4.1.0.** v4 records HTTP responses as Jest snapshots (not the old `__playbacks__/` cassettes) and intercepts requests via `@mswjs/interceptors`, which models responses with `fetch`/`Response` semantics.
- **Move `get-version` from `got` to native `fetch`, and drop the `got` dependency.** v4's interception is incompatible with `got`: a recorded response keeps its `content-encoding: gzip` header while the body is stored decoded, so `got` re-runs gzip decompression and fails with `incorrect header check`. `got` was only used here for two trivial GET requests (added in #211, before Node had a global `fetch`). `fetch` reads the intercepted body without re-decompressing; an explicit `response.ok` check preserves got's throw-on-error behaviour.
- **Split `getVersionObject` into `fetchVersions` + a pure `resolveVersion`.** v4 keys recordings by test name, so fetching `versions.json` once per `it.each` case would record the large response 9×. The test fetches the list once and runs every range assertion against it in memory, keeping the snapshot to a single copy.

## Jest runs as native ESM

Rather than transpiling jest-playback back to CommonJS, the suite now runs as native ESM:

- ts-jest `default-esm` preset + `extensionsToTreatAsEsm`; the `test` script runs with `--experimental-vm-modules`; tsconfig moves to `module: esnext` / `moduleResolution: bundler` (also resolving jest-playback's `exports`-only types), `target: es2020` (top-level `await`), and `isolatedModules`.
- `jest.mock` factories → `jest.unstable_mockModule` + dynamic `import`, with `@jest/globals` imported where `jest` is used. Modules reached through the CommonJS `actions-parsers` wrapper (`@actions/core`) keep `jest.mock`, which intercepts the CommonJS `require`.
- `os`/`@actions/exec` mocks moved inline (the auto-applied manual mock no longer works under ESM); the github context is mutated through a helper (ESM namespaces are read-only); `__dirname` derived from `import.meta.url`; `ts-invariant` imported by name.

The old `__playbacks__/` cassettes are deleted and the response re-recorded into the snapshot.

## Verification

- `yarn build` ✅ (clean ncc bundle, no parse placeholders)
- `CI=true yarn test` ✅ — 8 suites, 45 tests; snapshots replay from recordings (no network)
- `yarn lint` ✅, `tsc --noEmit` ✅


iwahbe:
This change is part of the following stack:

- #1434 ◀
    - #1435

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

tgummerer (approved):

